PR #19, opened 2021-02-13
feat(docker): configurable DNS server

sebbrandt87:
- DNS_1 and DNS_2 environment variable added

So now we are able to configure our prefered DNS servers through the `/etc/sysconfig/valheim-server` file if we want to override the defaults.


lloesche (approved):

lloesche:
LGTM! Thanks.

PR #21, opened 2021-02-13
Added the Windows Parameters "-nographics -batchmode" to the valheim-server script

Supportlik:
First of all: Great work!

I added the Windows Parameters "-nographics -batchmode" from the Steam Dedicated Server Package to the valheim-server script. 

This leads to 10% less CPU usage and less desync lag on my server.

### DIFF
--- a/valheim-server
+++ b/valheim-server
@@ -22,7 +22,7 @@ main() {
             sleep 10
         fi
     done
-    echo "Running Valheim Server - EXPERIMENTAL"
+    echo "Running Valheim Server"
     "$valheim_server" -nographics -batchmode -name "$SERVER_NAME" -port $SERVER_PORT -world "$WORLD_NAME" -password "$SERVER_PASS" -public $SERVER_PUBLIC &
     valheim_server_pid=$!
     wait
--- a/valheim-server
+++ b/valheim-server
@@ -23,7 +23,7 @@ main() {
         fi
     done
     echo "Running Valheim Server"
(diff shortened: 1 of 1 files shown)


lloesche:
Thank you, LGTM.

lloesche (approved):

InB4DevOps:
I updated my container to latest version but I don't see a change in CPU load at all. Neither in idle nor with players connected. Any idea?

LHolst:
I can confirm a drop of 20%ish . Before 4cpu at 120%, after updating I'm running below 100% .

Great work!

InB4DevOps:
I imagine the nographics switch only applies to systems that actually have a GPU. Since my NAS doesn't have a GPU that switch might not have an effect for me.

lerb:
I've noticed a massive CPU-usage drop and a pretty hefty RAM-usage drop too since this was introduced. Docker used to show around 45% CPU usage and 2,5GB RAM usage, it's now down to 15% CPU/1,8GB RAM. I'm hosting on a Synology DS218+

Great work!

lloesche:
@lerb that sounds more like you're experiencing the effects of server `v0.145.6` which had this in the changelog:
```
Dedicated server CPU usage fix (Framerate limiter)
```
Was released like two days ago or so and had that extreme CPU usage drop effect you're describing.

lerb:
> @lerb that sounds more like you're experiencing the effects of server `v0.145.6` which had this in the changelog:
> 
> ```
> Dedicated server CPU usage fix (Framerate limiter)
> ```
> 
> Was released like two days ago or so and had that extreme CPU usage drop effect you're describing.
> 
> See https://store.steampowered.com/news/app/892970/old_view/3059603718045184990 for more details.

Oh, I see! Silly me, thank you for the clarification.

PR #70, opened 2021-02-21
Update docker-compose.yaml

okazuki0919:


Toucan-Sam (commented):
@okazuki0919 This pull request adds YOUR server name and your server's password to the common docker-compose.yaml.
Probably not what you meant to do, but this should absolutely not be accepted and you should consider changing your server password. :)

lloesche:
Looks like an unintentional PR. Reopen with description if not.

Issue #63, opened 2021-02-19
Add support for Valheim Plus Mod


alec-hs:
Adding support for Valheim Plus mod via an ENV Variable would be great. Stuff probably needed to make this work:

- ENV Variable to enable/disbale
- New volume with Valheim Plus config 
- Removal of "validate" command to stop steam overwriting mod files

<https://github.com/nxPublic/ValheimPlus>

armanatz:
@lloesche I am actually working on this as we speak

kachunkachunk:
Woo, glad to hear (and see this submitted)! I think the big gotcha in my view surrounds the supervisor bits and loading BePinEx, but I think it'd be a bunch of hours/days for me to figure out, if I can even get to that point.

nxPublic:
Dawm you guys are great, thanks for taking the effort into it. ❤️ 

This is a great documentation and i will make sure to link it in the V+ repo as soon as its ready to.

Thank you!

armanatz:
@lloesche PR submitted for this. Let me know what you think

wedol:
See https://github.com/lloesche/valheim-server-docker/issues/72#issuecomment-783307733

lloesche:
If you guys would like to test ValheimPlus support you can use the dev tag of the Docker image `lloesche/valheim-server:dev`

Updated README is here https://github.com/lloesche/valheim-server-docker/tree/dev

Let us know in https://github.com/lloesche/valheim-server-docker/pull/91 if everything works.

Thanks to all testers!

lloesche:
@nxPublic I'd very much like your input on #91 
I tried using the included `start_game_bepinex.sh` the issue however is that it has no process control meaning if I kill the script the server keeps on running. One way of solving this would be to `exec` the server and just not reset LD_LIBRARY_PATH. Or track the server process and trap SIGINT/SIGTERM and bubble them up to the server PID.
Anyhow, I just took the contents and added them to the `valheim-server` start script.

NoobTaco:
Testing now. One question, what about other mods? For example, the Craft from All Containers mod? If I drop the dll in the data/plus/plugins directory, will it be overwritten on update?

lloesche:
@NoobTaco it'll work however anything in the data directory is regenerated when updates happen. We could add something that allows to add additional mods. That's a separate issue though.

lloesche:
ValheimPlus support has been merged into main

Issue #20, opened 2021-02-13
Specify user and group ID


coolmule0:
Is it possible to change the owning user and group of the container so root does not own everything? 

This helps with mounted volumes not be root access level

Supportlik:
I will add this feature this weekend.

ericlathrop:
I tried running this container as non-root, and got an error about supervisord being unable to drop privileges.

Do-while-Bacon:
Just adding in that I would really like to have this feature. I feel that it is good security practice to isolate containers (and associated volumes) under their own service accounts.

lloesche:
I would love this feature. Here's a couple of things to consider for an implementation:

This image is being used by a lot of NAS users who might not be familiar with UIDs and changing permissions. I do not want to break the image for them. For instance on a Synology when starting the server on a directory that hasn't been properly set up the container might see permissions `000` on all the world files copied to the /config/worlds directory with some unknown UID/GID as owner. Solving this is how the `WORLDS_DIRECTORY_PERMISSIONS` and `WORLDS_FILES_PERMISSIONS` env vars were born.

For tech savvy users it is already possible to run the container in a different User namespace using subordinate UID/GIDs so UID 0 is not mapped to the hosts root account anyways and therefor changing the UID within the container has little effect on the overall security of the system. For anyone concerned with security of the host system this is likely a must-do regardless.

There are multiple levels of implementation that we could look at.

One would be building the container and running supervised and the valheim-server/-updater/-backup scripts as root but running the actual server binary as non-root.
This would still allow the container to change permissions on files like described above but add some security to the one component that is exposed to the Internet and therefor vulnerable to remote exploits.
This would also allow to supply a UID/GID via environment variables and change the server permissions to those IDs on startup.

The other would be to drop privileges right in the Dockerfile. Essentially the image would be built as root but then permissions are changed to UID `1001` and right before the `ENTRYPOINT` we would have `USER 1001`.
This is how Bitnami builds all their images. It has the obvious advantage that none of the processes inside the container run as root but the obvious disadvantage that anyone volume mounting external world files into the container must make sure that UIDs inside and outside the container match up. Also the UID/GID can not be changed at runtime and are set to a fixed value in the Dockerfile at build time. This I feel is prohibitively complex for the NAS user that just wants to play some Valheim with their friends but does not have the required Linux knowledge to make this work for them.

[...]

PR #76, opened 2021-02-21
Implement Valheim Plus mod support

armanatz:
Takes care of #63

I apologize if my Bash code isn't elegant but this is the best I could do with my limited knowledge. In general, the steps added are as follows:

- Created an ENV variable that allows the user to specify whether V+ is enabled (by default disabled).
- If the mod is enabled, during the `update()` function call, we first get an md5 checksum list of all the unmodified server files to keep as reference.
- Then we call the new `check_valheimplus()` function to see if we need to install/reinstall the mod.
- Inside `check_valheimplus()` we do either a check to see if the mod exists in the first place (this is usually on first start of container) of if the md5 checksums of the server files don't match the files found in the mod (this is for cases where Steam overrides the files).
- Inside the `install_valheimplus()` function, we go ahead and download and install the mod in a volume specifically for it found at `/opt/valheimplus` (the reason we persist this volume is for checksum data as well as keeping the latest zip file downloaded there plus the unzipped files from the mod).
- We then `rsync` all the files in the mod volume to both `/opt/valheim_dl` and `/opt/valheim`
- I have also implemented a new `run_bepinex.sh` script that should take care of handling our server starting function better.
- The `update_valheimplus()` function checks if the md5 checksums of the remote zip file and the local zip file differ, in which case it will run `install_valheimplus()` again to update the mod.

The only bit of concern here is that the `valheim_plus.cfg` file needs to have its ownership changed to reflect the UID and GID of the user running the container. This is so that users can easily modify the file without needing root access. However, I think if #20 was implemented, we could switch over to that


armanatz:
OK, so I noticed some issues with the way `rsync` is handling some files. BepInEx, once run actually creates some new files and directories that are only available in `/opt/valheim` and not in `/opt/valheim_dl` which causes the mod to keep reinstalling. Let me look into fixing that as well

armanatz:
@lloesche How averse are you to using md5 checksums to check if the server files were updated against `/opt/valheim` instead of using `grep '^[*>]' "$logfile" > /dev/null 2>&1` as the output for that line is a bit unreliable

lloesche:
@armanatz thank you for your PR. I'll review it tomorrow. Can you give some details how the rsync output is unreliable? A line beginning with a `*` means a file was removed and a line beginning with a `>` means a file was added/replaced.

armanatz (commented):
@lloesche So [this commit's](https://github.com/lloesche/valheim-server-docker/pull/76/commits/25c9a390fbe8b33767c8cc375ab2cb7c2b8d4ff0) story began as me trying to ensure `rsync` doesn't mess with the Valheim Plus mod files.

That led me down a rabbit hole of refactoring and re-implementing some of the update logic for the Docker container which also led me to changing the behavior you had implemented in #15.

The new logic is as follows:

1. On container fresh start, we install the server files to the `/opt/valheim` folder.
2. Then, if Valheim Plus is enabled, we install that and take the relevant steps to ensure the mod works.
3. Upon every update check, we first get the local manifest ID of the Valheim Server package from Steam (this is found in the folder `/opt/valheim/steamapps/appmanifest_896660.acf`).
4. We use this ID to then check against the remote ID found on Steam's own DB.
5. If the manifest ID differs between local and remote, the update is triggered and things should theoretically work as expected (this is the only part of the script that I can't test for obvious reasons).
6. In order to ensure everything works correctly, I have disabled passing `-validate` as a `STEAMCMD_ARGS` but rather use `FORCE_VALIDATE` to allow users to choose if they want to validate their files.
7. Because of your comments in #7, the `install_valheim()` function still runs with the `-validate` flag just in case that was what was causing the issue of updates not registering in the container.

This whole thing has the benefit of ensuring:

- We ensure we check against Steam that there is actually an available update
- We reduce the container size back to its original size before #15 was implemented
- We ensure that the V+ mod files don't get accidentally deleted or overwritten during update checks which was what was happening before

Let me know what you think of this approach.

PS. There is a line in my commit that I want to explain as it seems redundant. Specifically [this one](https://github.com/lloesche/valheim-server-docker/pull/76/commits/25c9a390fbe8b33767c8cc375ab2cb7c2b8d4ff0#diff-efa67020bc87f32bbccdd6f62a754ae3246bf9fe0c4cc0d268af640e5d94716eR70).

From doing some research, I read that there can be an issue when pulling app info from Steam using `steamcmd` if you don't first run `+app_update`.

The issue is that the information in the app info data dump might still be outdated and so you need to force `steamcmd` to actually do an update in order to refresh its cache of info.

However, this may not actually be the case and because there is no Valheim Server update for me to check, I am not sure of the necessity of this line. It could be that the entire thing is unnecessary and redundant.

Dockerfile
```diff
@@ -11,18 +11,21 @@ RUN dpkg --add-architecture i386 \
...
         libsdl2-2.0-0:i386 \
+        libc6-dev \
         ca-certificates \
         supervisor \
         procps \
         locales \
         unzip \
         zip \
         rsync \
+        wget \
```
lloesche, line 22: When you rebase on the current master you get wget from busybox. Added it for syslog and crond.

README.md
```diff
@@ -1,30 +1,35 @@
...
     -p 2456-2458:2456-2458/udp \
-    -v $HOME/valheim-server-config:/config \
+    -v $HOME/valheim-server/config:/config \
+    -v $HOME/valheim-server/files:/opt/valheim_dl \
     -e SERVER_NAME="My Server" \
-    -e WORLD_NAME="Neotopia" \
     -e SERVER_PASS="secret" \
+    -e WORLD_NAME="Neotopia" \
+    -e PUBLIC=1 \
+    -e VALHEIM_PLUS_ENABLED=true
```
lloesche, line 32: We should disable by default as Valheim Plus requires clients to have the mod installed as well (except for server only settings). People should intentionally enable the mod.

valheim-server
```diff
@@ -1,18 +1,17 @@
...
 SERVER_NAME=${SERVER_NAME:-My Server}
 SERVER_PORT=${SERVER_PORT:-2456}
 WORLD_NAME=${WORLD_NAME:-Dedicated}
 SERVER_PASS=${SERVER_PASS:-secret}
 SERVER_PUBLIC=${SERVER_PUBLIC:-1}
+VALHEIM_PLUS_ENABLED=${VALHEIM_PLUS_ENABLED:-false}
 timeout=20
 kill_signal=TERM
 
-export SteamAppId=892970
```
lloesche, line 13: These are common settings they should still be up here.

valheim-server
```diff
@@ -23,8 +22,27 @@ main() {
...
+        lib_postfix="so"
+        export LD_LIBRARY_PATH=./linux64:$LD_LIBRARY_PATH
+        doorstop_libname=libdoorstop_${arch}.${lib_postfix}
+        export LD_LIBRARY_PATH="${doorstop_libs}":${LD_LIBRARY_PATH}
+        export LD_PRELOAD=$doorstop_libname:$LD_PRELOAD
+        export DYLD_LIBRARY_PATH="${doorstop_libs}"
+        export DYLD_INSERT_LIBRARIES="${doorstop_libs}/$doorstop_libname"
+        export templdpath=$LD_LIBRARY_PATH
+        export SteamAppId=892970
+    else
```
lloesche, line 39: This entire else case is unnecessary. The values are the same for Valheim vanilla and Valheim Plus. V+ just extends the LD_LIBRARY_PATH. Also there is no need to restore the old LD_LIBRARY_PATH as this script is never run from a shell where restoring the original state would be of interest.

lloesche (changes requested):

valheim-updater
```diff
@@ -5,39 +5,81 @@ CONFIG_DIRECTORY_PERMISSIONS=${CONFIG_DIRECTORY_PERMISSIONS:-755}
...
+    unzip -qo /opt/valheimplus/UnixServer.zip -x changelog.txt run_bepinex.sh -d /opt/valheimplus
+    if [ -f /opt/valheim/BepInEx/config/valheim_plus.cfg ]; then
+        cp -a /opt/valheim/BepInEx/config/valheim_plus.cfg /opt/valheimplus/BepInEx/config/valheim_plus.cfg
+    fi
+    chmod 666 /opt/valheimplus/BepInEx/config/valheim_plus.cfg
+    rsync -a --itemize-changes --exclude UnixServer.zip /opt/valheimplus/ /opt/valheim
+}
+
+update_valheim_plus() {
+    remote_md5="$(curl -sL https://github.com/nxPublic/ValheimPlus/releases/latest/download/UnixServer.zip | md5sum | cut -d ' ' -f 1)"
```
lloesche, line 41: We can add `jq` and then the download URL size and update time can be retrieved using e.g.
```
curl -sL https://api.github.com/repos/nxPublic/ValheimPlus/releases/latest | jq -r '.assets[] | select(.name == "UnixServer.zip") | .browser_download_url,.size,.updated_at'
```
We can store the resulting update date together with the ZIP, for instance in a UnixServer.zip.updated_at file or something like that. Then on update check we just do that one API call instead of downloading the ZIP file and compare date and size.

valheim-updater
```diff
@@ -5,39 +5,81 @@ CONFIG_DIRECTORY_PERMISSIONS=${CONFIG_DIRECTORY_PERMISSIONS:-755}
...
+    echo "Valheim Server was updated - restarting"
+    supervisorctl restart valheim-server
+}
+
+update_valheim() {
+    if [ ! -d /opt/valheim/valheim_server_Data ] || [ $FORCE_VALIDATE = true ]; then
+        echo "Installing Valheim Server"
+        args=""
+        if [ $FORCE_VALIDATE = true ]; then
+            args=-validate
```
lloesche, line 64: I believe the arg is just `validate` not `-validate`.

valheim-updater
```diff
@@ -5,39 +5,81 @@ CONFIG_DIRECTORY_PERMISSIONS=${CONFIG_DIRECTORY_PERMISSIONS:-755}
...
+            FORCE_VALIDATE=false
+        fi
+        install_valheim $args
     else
-        echo "Valheim Server is already the latest version"
-        if [ $just_started = true ]; then
-            supervisorctl start valheim-server
+        local_version=$(sed -n '29p' /opt/valheim/steamapps/appmanifest_896660.acf | xargs | cut -d' ' -f2-)
```
lloesche, line 69: Seems rather fragile to go by line number. Can we parse the value in a more reliable way?

valheim-updater
```diff
@@ -5,39 +5,81 @@ CONFIG_DIRECTORY_PERMISSIONS=${CONFIG_DIRECTORY_PERMISSIONS:-755}
...
+    local_md5="$(md5sum /opt/valheimplus/UnixServer.zip | cut -d ' ' -f 1)"
+    if [ "$remote_md5" != "$local_md5" ]; then
+        install_valheim_plus
+    else
+        echo "Valheim Plus is already the latest version"
+    fi
+}
+
+install_valheim() {
+    ./steamcmd.sh +login anonymous +force_install_dir /opt/valheim +app_update 896660 $1 $STEAMCMD_ARGS +quit
```
lloesche, line 51: What happens to existing users who extended STEAMCMD_ARGS and already have validate plus whatever else they were using in here? Do we have to remove validate from their STEAMCMD_ARGS in order for this mod to work? Should we tell them in the README that validate in STEAMCMD_ARGS is not compatible with ValheimPlus mod?

README.md
armanatz: By default, the container will not run with Valheim Plus unless explicitly asked to. I put this in the README as an example for people who want to just quickly get started but I can see how it can be confusing. Will remove it

valheim-updater
armanatz: Ah, I completely forgot about checking against Github's API. Yeah, your suggestion sounds better to me

valheim-updater
```diff
@@ -5,39 +5,81 @@ CONFIG_DIRECTORY_PERMISSIONS=${CONFIG_DIRECTORY_PERMISSIONS:-755}
...
+            FORCE_VALIDATE=false
+        fi
+        install_valheim $args
     else
-        echo "Valheim Server is already the latest version"
-        if [ $just_started = true ]; then
-            supervisorctl start valheim-server
+        local_version=$(sed -n '29p' /opt/valheim/steamapps/appmanifest_896660.acf | xargs | cut -d' ' -f2-)
+        ./steamcmd.sh +login anonymous +force_install_dir /opt/valheim +app_update 896660 $STEAMCMD_ARGS +quit
+        remote_version=$(./steamcmd.sh +login anonymous +app_info_update 1 +app_info_print 896660 +quit | sed -n '/"896661"/,/"public"/ { /"896661"/!p; }' | sed -n '9p' | xargs | cut -d' ' -f2-)
```
lloesche, line 71: Same concern as before, this is very fragile code. We'll need proper parsing for the format. Seems to be the same as the acf file above.

valheim-updater
armanatz: Actually, both work fine.

valheim-server
armanatz: Yeah, I just wanted to insert the original code found in the `run_bepinex.sh` script and go from there. However, I will refactor this section to be less redundant

lloesche:
@armanatz I very much like the approach of checking steam for the current build version and comparing it with what we have local. I just looked over the PR and added some comments where I saw some obvious things. I like the overall design, but I'm not a fan of relying on line numbers in remote generated content. That'll be the first code to break.

Regarding container size, I'm not at all concerned with that. If the container image is 2GB larger at runtime but therefor we have cleanly separated download and install directories that's fine by me. Disk space is cheap and plentiful.

If we can't get the steam version parsing code to be more robust I'm also fine with keeping the current /opt/valheim_dl and /opt/valheim folders and (for plus users) creating yet another /opt/valheim_plus directory which would be a merger of /opt/valheim and UnixServer.zip. valheim-server can then decide whether it starts the vanilla server in /opt/valheim or the plus server in /opt/valheim_plus. The plus would also be that users can switch between both versions on the fly.

We want to be sure that we
- do not break anything for existing users
- make the code robust to changes on the IronGate, ValheimPlus and Steam side of things

lloesche:
Just did a quick search and https://github.com/leovp/steamfiles seems to have a very robust acf and appinfo parser implementation. Is that something you could tackle? If not I can invest some time on Wednesday.

README.md
lloesche: I figured, but I'd just change true to false. That way people see that the container supports V+ but don't accidentally enable it.

armanatz:
> I like the overall design, but I'm not a fan of relying on line numbers in remote generated content. That'll be the first code to break.

Yeah, I will look into parsing things a bit better. To be honest, my knowledge with `sed` is abysmally tiny.

> Regarding container size, I'm not at all concerned with that. If the container image is 2GB larger at runtime but therefor we have cleanly separated download and install directories that's fine by me. Disk space is cheap and plentiful.

True, however there are some people who'd like to run the server on an RPi and I was thinking any space savings would be welcomed for them. I think if we can keep the behavior the way it is currently, it would also make understanding the folder structure of the container as well as future maintenance a bit easier as well since we are only dealing with a single game folder.

lloesche:
I just played around with that Python library. Works quite well. Here's my rough prototype for a acf to json converter:
```
root@2f403fbaf80b:/opt/valheim_dl/steamapps# cat ~/acf2json.py
import sys
import select
import json
from steamfiles import acf

if select.select([sys.stdin,],[],[],0.0)[0]:
    acf_content = sys.stdin.read()
    data = acf.loads(acf_content)
    print(json.dumps(data))
```

Will produce output like:
```
root@2f403fbaf80b:/opt/valheim_dl/steamapps# python3 ~/acf2json.py < appmanifest_896660.acf | jq .
{
  "AppState": {
    "appid": "896660",
    "Universe": "1",
    "name": "Valheim Dedicated Server",
    "StateFlags": "4",
    "installdir": "Valheim dedicated server",
    "LastUpdated": "1613959631",
    "UpdateResult": "0",
    "SizeOnDisk": "1045271601",
    "buildid": "6246034",
    "LastOwner": "76561198828919924",
    "BytesToDownload": "0",
    "BytesDownloaded": "0",
    "BytesToStage": "0",
    "BytesStaged": "0",
    "AutoUpdateBehavior": "0",
    "AllowOtherDownloadsWhileRunning": "0",
    "ScheduledAutoUpdate": "0",
    "InstalledDepots": {
      "1006": {
        "manifest": "6688153055340488873",
        "size": "59862244"
      },
      "896661": {
        "manifest": "521795651741005384",
        "size": "985409357"
      }
    },
    "UserConfig": {}
  }
}
```
I think this can be useful. We can either process it in Python or just use the lib to convert to json and then do further selection with jq. I'm fine with either. Should be pretty robust.


Although I'm not sure if this is worth installing an entire python stack. At that point we could just leave the existing rsync code which we know works reliable and does not require any metadata parsing and create a third /opt/valheim_plus directory.

valheim-updater
armanatz: How about doing this line like this:

```
awk '/"896661"/,/"manifest"/{if($1=="\"manifest\"")print $2;}' /opt/valheim/steamapps/appmanifest_896660.acf | xargs
```

So basically we use `awk` to extract everything between `"896661"` and `"manifest"`, then we check if the first column matches `"manifest"` and extract the second column. Finally we pass it to `xargs` to strip extraneous characters and return just the raw ID

valheim-updater
armanatz: This is also fixed by doing the same thing as above. Something like this:

```
./steamcmd.sh +login anonymous +app_info_update 1 +app_info_print 896660 +quit | awk '/"896661"/,/"public"/{if($1=="\"manifest\"")print $2;}' | xargs
```

So same deal but just that we are passing the output of `steamcmd` to `awk`

valheim-updater
armanatz: With regards to this, I think it is better to strip `-validate` from `STEAMCMD_ARGS` and just redirect that behavior to `FORCE_VALIDATE` instead. This has the benefit of making `-validate` a separate and intentional flag while not breaking current user `STEAMCMD_ARGS`. Although I don't believe for Valheim Server there are any additional arguments that can be passed to `steamcmd` as they don't have a publicly available beta branch

armanatz:
@lloesche If all looks good, I think my recent changes should suffice with regards to merging this PR. Let me know your thoughts :)

README.md
```diff
@@ -1,31 +1,35 @@
...
-
+Valheim Server in a Docker Container (with optional [Valheim Plus](https://github.com/nxPublic/ValheimPlus) mod support)  
 
 # Basic Docker Usage
 
 The name of the Docker image is `lloesche/valheim-server`.
 
 Volume mount the server config directory to `/config` within the Docker container.
 
+If you would also like to enable Valheim Plus, please also volume mount the server files directory `/opt/valheim_dl` so that you can modify the mod's config file.
```
lloesche, line 12: This should be /opt/valheim

lloesche, line 12: Should this be /opt/valheim_dl?

README.md
```diff
@@ -1,31 +1,35 @@
...
 ```
-$ mkdir -p $HOME/valheim-server-config/worlds
+$ mkdir -p $HOME/valheim-server/config/worlds
 # copy existing world
 $ docker run -d \
     --name valheim-server \
     -p 2456-2458:2456-2458/udp \
-    -v $HOME/valheim-server-config:/config \
+    -v $HOME/valheim-server/config:/config \
+    -v $HOME/valheim-server/files:/opt/valheim_dl \
```
lloesche, line 27: Same question here. I thought we removed /opt/valheim_dl with this PR

supervisord.conf
```diff
@@ -19,40 +19,40 @@ password=dummy
 [rpcinterface:supervisor]
 supervisor.rpcinterface_factory=supervisor.rpcinterface:make_main_rpcinterface
 
-[program:syslogd]
```
lloesche, line 22: This file shouldn't be touched by this PR. I believe you just need to rebase on the latest main branch.

valheim-bootstrap
```diff
@@ -1,5 +1,15 @@
 #!/bin/bash
 RESTART_CRON=${RESTART_CRON-0 5 * * *}
+TZ=${TZ:-Etc/UTC}
```
lloesche, line 3: Same comment as above, this file shouldn't be part of this PR?! Did you git rebase on current main?

lloesche:
@armanatz lgtm 👍 just had two questions about `/opt/valheim_dl` still in the README and was confused about two files that shouldn't be part of the PR. Likely just a quick rebasing issue.

armanatz:
> @armanatz lgtm 👍 just had two questions about `/opt/valheim_dl` still in the README...

Woops, sorry. Forgot to make the changes to the README. It is fixed now.

> and was confused about two files that shouldn't be part of the PR. Likely just a quick rebasing issue.

That should be fixed as well. Was an issue on my part.

lloesche (approved):

lloesche:
merged, thank you! @armanatz

armanatz:
@lloesche Oh dang, there was a new update to V+ that changed its structure. I was going to message saying not to merge yet. I am doing a fix for it as we speak. Will submit a new PR then

lloesche:
@armanatz Hey, I just had to revert the PR because my server was restarting every 15 minutes on every update check.
Also valheim plus was enabled by default in the updater.

lloesche:
We'll have to redo this PR with your latest changes and also figure out why the updater is broken and restarts the server every 15 minutes with the changes you made. This time lets create it against the `dev` branch so we can test the image before merging into main.

armanatz:
> @armanatz Hey, I just had to revert the PR because my server was restarting every 15 minutes on every update check.

That's odd, it doesn't happen for me. Let me recheck and see again. Maybe I missed something.

> Also valheim plus was enabled by default in the updater.

That's weird. This wasn't from my side. Will double check.

> We'll have to redo this PR with your latest changes and also figure out why the updater is broken and restarts the server every 15 minutes with the changes you made. This time lets create it against the dev branch so we can test the image before merging into main.

Yeap, on it. Will do changes against `dev` branch

Hydeo:
Thank you two for your hardwork.
Once we have something on dev, I'll do my best to help with the testing

lloesche:
You had the code
```
VALHEIM_PLUS_ENABLED=${VALHEIM_PLUS_ENABLED:-true}
```
twice, once in -server once in -updater. In server it defaulted to `false` in updater it defaulted to `true`.

lloesche:
We should introduce a "default-vars" and set all the defaults in there and then include that inside of all of our scripts. That way we don't have to maintain the same default in two places.

armanatz:
> 
> 
> You had the code
> 
> ```
> VALHEIM_PLUS_ENABLED=${VALHEIM_PLUS_ENABLED:-true}
> ```
> 
> twice, once in -server once in -updater. In server it defaulted to `false` in updater it defaulted to `true`.

Yea, might have been some rogue commit I mistakenly pushed. My bad. Will ensure to double check. Lack of sleep does that to you haha.

> 
> 
> We should introduce a "default-vars" and set all the defaults in there and then include that inside of all of our scripts. That way we don't have to maintain the same default in two places.
> 
> I'll do that quick.

Yea, that sounds like a good idea. Thanks!

lloesche:
No worries, we caught it before it effected any users so all is good. I just pushed https://github.com/lloesche/valheim-server-docker/commit/704d05009dc468a19b30b237bb3d3aaa696a346c which creates a file `defaults` where we can maintain all our default env vars.

I was thinking to avoid any issues with server updates we should keep the current rsync update process instead of querying steam. It's just more reliable in that we only restart the server if files have actually changed. What do you think about the following structure:

```
/opt/valheim/dl/server   # here we let steamcmd download the dedicated server unmodified
/opt/valheim/dl/plus     # here we download and extract UnixServer.zip of ValheimPlus and store the updated_at date info
/opt/valheim/server      # here we rsync the vanilla server from /opt/valheim/dl/server and use the current restart logic
/opt/valheim/plus        # here we rsync with `--link-dest` (creating hard links) from /opt/valheim/server and /opt/valheim/dl/plus to create a ValheimPlus distribution
```

This would also be backwards compatible with folks who already have a /opt/valheim volume mount.

By using hard links in the last step we don't consume any more disk space than we do now (although as mentioned I'm not too concerned with +/- 1GB of disk space usage).

I also feel this makes every step of adding ValheimPlus more robust. If ValheimPlus for whatever reason breaks users can just turn it off and we will start Valheim from the unmodified `/opt/valheim/server` vanilla location.

lloesche:
@armanatz  By the way I just invited you as a collaborator for this repository. Let's create a new valheimplus branch and do the work in there. That way we can more easily collaborate on it.
Dockerhub is also set up to automatically build any branches we create as new tags. So a valheimplus branch would become available as `lloesche/valheim-server:valheimplus`.
That'll allow us better testing.

Hydeo:
> I also feel this makes every step of adding ValheimPlus more robust. If ValheimPlus for whatever reason breaks users can just turn it off and we will start Valheim from the unmodified `/opt/valheim/server` vanilla location.

That's is defo a important point. Modding on Valheim is, as far as I understand it, quite hacky so having a easy/quick way to fall back seems right.

>Dockerhub is also set up to automatically build any branches we create as new tags. So a valheimplus branch would become available as lloesche/valheim-server:valheimplus.
>That'll allow us better testing.

Indeed. Don't hesitate to ping me once you have something there, I have a few players that are willing to test the mod, I should be able to get you some feedback quite fast.

lloesche:
@armanatz let me know what you think of https://github.com/lloesche/valheim-server-docker/pull/88

@Hydeo thank you! Will definitely ping you once a PR is ready to test!

armanatz:
> I also feel this makes every step of adding ValheimPlus more robust. If ValheimPlus for whatever reason breaks users can just turn it off and we will start Valheim from the unmodified /opt/valheim/server vanilla location.

@lloesche Actually V+ changed the way it handles things with BepInEx. So before 0.8.5, they used to ship a folder called `valheim_server_Data/Managed` which had unstripped Unity DLLs in it. This folder would modify the `valheim_server_Data/Managed` folder downloaded via Steam and overwrite things.

With 0.8.5, they changed the behavior by instead having a `unstripped_corlib` folder shipped with new releases. This folder has the same files as `valheim_server_Data/Managed` that was shipped before. However, now they load this folder into Doorstop instead.

What this effectively means is that none of the actual files on the server are modified. This also means that Steam won't always need to revalidate the game files if V+ is installed.

So what I am trying to say is that we don't have to worry about V+ breaking things. Basically, if the user disables `VALHEIM_PLUS_ENABLED`, then we just load things in the vanilla way and they are good to go

armanatz:
> I was thinking to avoid any issues with server updates we should keep the current rsync update process instead of querying steam. It's just more reliable in that we only restart the server if files have actually changed.

I don't know if `rsync` is the right tool though over here. It also can cause some issues if in the future IronGate decides to create temp files or cache files directly in the server folder. Or in the case of BepInEx, it actually does create a cache directory on server start, therefore, anytime the update check runs, it will restart the server if you are running V+.

By saying that we run `rsync` whenever there is a modification to a file and then restarting the server, we may end up restarting unnecessarily when there actually were no updates to sync.

By using Steam as our single source of truth for updates, then we can guarantee that the information is correct as Steam sees it. If when Steam notifies that an update is there, then only we initiate a restart.

lloesche:
> It also can cause some issues if in the future IronGate decides to create temp files or cache files

That's just an issue of not supplying the `--delete` argument and/or not checking for lines starting with `*`. If we only check for updated files that were transferred to the destination `>` then we should be good.
And there is no conceivable way where the game would modify its own files that were downloaded from Steam because then Steam would always report them as corrupt. So I feel this is a pretty robust method of checking for changes.

lloesche:
> If when Steam notifies that an update is there, then only we initiate a restart.
I would agree if we had programmatic access to the Steam API. Calling steamcmd and parsing the output with a number of sed/awk/trs or even via the acf library seems less than ideal. It's a bit like scraping websites for information and hoping they don't change the format.

lloesche:
@Hydeo as promised #91 for you to play around with 🙂 
Docker image is available as `lloesche / valheim-server:dev`

Hydeo:
Awesome !
I'll play with it today, I'll come back to you as soon I as have :)

PR #87, opened 2021-02-23
simplistic capability to enable aws and define an s3 path

ryphon:
very simple. want to get cleaning of old backups off of s3 working.

adds bloat with awscli in the container, and relies on being in an ec2 to be able to pull the metadata credentials, but I'm running this whole thing on a bare ec2 as a service in an autoscaling group.

I do a few commands at startup to pull the server from bucket, then start up the docker pointing at the files I pulled from s3.

The idea would be that it'd pull the latest backup. Still need to figure out how to do a quick 'shutdown-save-s3-backup' kinda thing, but for now this is better than nothing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ RUN dpkg --add-architecture i386 \
         unzip \
         zip \
         rsync \
+        awscli \
     && echo 'LANG="en_US.UTF-8"' > /etc/default/locale \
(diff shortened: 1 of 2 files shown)


lloesche:
I suggest you create a second container that you deploy in the same pod or same container host and also has access to the $BACKUPS_DIRECTORY volume and do the sync in there. That seems cleaner.

PR #91, opened 2021-02-24
Initial ValheimPlus support

lloesche:
@armanatz I had some time tonight and since you didn't re-open your PR for me to review I sat down and added support for #63 

It seems to throw a lot of errors upon startup along the lines of
```
Fallback handler could not load library /opt/valheim/plus/valheim_server_Data/Mono/�l�:V/mono/aot-cache/amd64/UnityEngine.PhysicsModule.dll.so
```

but the server starts up and the mod seems to be loaded. I changed the
```
[Server]
dataRate=600
```
and my client showed receives in excess of 200 kb/s. So I'm assuming this is working. Didn't try anything mod wise client side though.

It'll create/expect the config in `/config/valheimplus`.

The Docker image is available as `lloesche/valheim-server:dev`


valheim-plus-updater
```diff
@@ -0,0 +1,88 @@
...
+    check_for_update
+}
+
+create_directory_structure() {
+    mkdir -p "$vp_download_path" "$vp_install_path"
+}
+
+check_for_update() {
+    cd "$vp_download_path"
+    local api_response=$(curl -sL https://api.github.com/repos/nxPublic/ValheimPlus/releases/latest | jq -r ".assets[] | select(.name == \"$vp_zipfile\")")
```
armanatz, line 18: Just a comment on this, GitHub API is rate limited for unauthenticated requests. Realistically, no one should hit it as the limit is 60 requests an hour but maybe something we should add in the README under the Updates section?

valheim-plus-updater
```diff
@@ -0,0 +1,88 @@
...
+    curl -L -o "$download_path" "$download_url" \
+        && echo "$updated_at" > "$download_path.updated_at"
+}
+
+extract_valheim_plus() {
+    cd "$vp_download_path"
+    echo "Extracting downloaded ValheimPlus ZIP archive"
+    rm -rf extracted
+    mkdir -p extracted
+    unzip -d extracted/ $vp_zipfile
```
armanatz, line 85: You should consider only extracting the folders `BepInEx`, `unzipped_corlib`, and `doorstop_libs` as the rest of the files are not necessary

armanatz:
> It seems to throw a lot of errors upon startup along the lines of

These errors are normal according to the mod's creator although I wish they weren't there. Probably has something to do with BepInEx or Doorstop

valheim-plus-updater
```diff
@@ -0,0 +1,88 @@
...
+merge_valheim_plus() {
+    rm -rf "$vp_install_path.tmp" "$vp_install_path.old"
+    mkdir -p "$vp_install_path.tmp"
+    rsync -a --itemize-changes --exclude server_exit.drp --exclude steamapps "$valheim_download_path/" "$vp_install_path.tmp"
+    rsync -a --itemize-changes "$vp_download_path/extracted/" "$vp_install_path.tmp"
+    if [ ! -d "/config/valheimplus" ]; then
+        cp -a "$vp_install_path.tmp/BepInEx/config" "/config/valheimplus"
+    fi
+    rm -rf "$vp_install_path.tmp/BepInEx/config"
+    chmod +x "$vp_install_path.tmp/start_game_bepinex.sh"
```
armanatz, line 60: This script is probably not necessary. I think it is only for client startup and not server startup

valheim-plus-updater
```diff
@@ -0,0 +1,88 @@
...
+    ln -s /config/valheimplus "$vp_install_path.tmp/BepInEx/config"
```
armanatz, line 61: Should we `chmod` this directory so that it can be writeable by the user? Otherwise, only `root` can write to the `valheim_plus.cfg` file

valheim-plus-updater
lloesche: Great idea! Will do.

valheim-plus-updater
lloesche: Ah right, good catch! I don't even use it anymore. I tried to but it's flawed. They don't `exec` the last line and don't do process control in it. So when I tried to restart the server it just killed the bash instead of the actual server process, so I decided to do it all inside valheim-server and skip any of their shell scripts.

valheim-plus-updater
lloesche: I was contemplating introducing new env variables for those config file permissions. But then figured I'd also want to `chown` to another UID potentially. And then I figured if it was an issue we'd already have it for the existing `/config` files which also get created as UID `0` if the user doesn't supply them. Like, if the user volume mounts existing config it won't be touched. So I wasn't sure what to do here. But maybe adding the ability to `chmod` the files is a good start.

valheim-plus-updater
lloesche: Hmm in that case I would likely want to extract everything but then only sync the relevant folders in the merge process. Sorta like we do with the valheim server Steam download where some stuff gets excluded. At the same time, if they introduce new directories we don't have to change the code and the extra files don't seem to hurt. So I'm torn on this one.

lloesche: Is there anything you would want to explicitly exclude? Maybe the .sh scripts?

valheim-plus-updater
lloesche: Okay, introduced `VALHEIM_PLUS_CONFIG_DIRECTORY_PERMISSIONS` and `VALHEIM_PLUS_CONFIG_FILE_PERMISSIONS`

Hydeo:
Tested with one other player, worked like a charm :) 
I'll do some more fine tuning of the cfg file tonight and play for real, i'll report if anything. 
The mod is going to be updated automatically, you might want to say so on the readme.

lloesche:
@Hydeo great point, I'll update the README.

Hydeo:
Just a detail and not really V+ related : what do you think about simplifying the permission arguments for the files / folder ? 
A global option to give the same right to the whole folder would reduce the repetition. 
If I am correct I need to pass 6 arguments if I want the whole folder to have the same permission with V+.

lloesche:
@Hydeo great suggestion. I changed it so now all you have to set is a single variable `PERMISSIONS_UMASK=022` 🙂 
All the old env variables still work though. If not set they are being derived from the umask. Allowing for fine grained control for those who need it.

lloesche:
We've had this running for 24h now on our server and not seen any issues. Will merge tomorrow evening (UTC) if no objections/issues come up.

Hydeo:
Played a few hours, no issue to report. 

Thank you @armanatz  and @lloesche for implementing this.

valheim-plus-updater
NullFragment: In the same vein, they have renamed the start_server_bepinex.sh script at least once based on the documentation. Also incorporating all of the exports from that shell script into the valheim-server code means that you now have to maintain and mirror any changes that are pushed to that script.

Wouldn't it just be safer to keep the entire mod, then just find and remove the line that calls `./valheim_server`, then the run_server method could just run the mod's shell script to export the needed values, and launch the server with the config passed into the docker?

Nanolab53:
Just a quick comment to let you know that V+ just get upgraded to version [0.9](https://github.com/valheimPlus/ValheimPlus/releases/tag/0.9)

Hydeo:
And we also got an Valheim update. 
As far as this docker image is concern it seems fine, the game and the mod has updated and once I updated my client I could connect without any issues. 
I'm currently testing, some things (e.g share position by default) are not working anymore but that's more on V+ side, the Valheim changed a few things map and network wise. 

I couldn't find what was the version of V+ installed on the server, is this info in a file not exposed by docker or it doesn't exist ?

Hydeo:
They seems to have some known issue with their 0.9 release. 

`We are aware of the packet size buffer issues that are being reported and are investigating them.`

That probably not necessary right now but how about an option to give the version we want to pull from V+ ? If the new version breaks in a bad way we don't have a way to rollback to a stable one today.

lloesche:
@Nanolab53 thanks, I built this container so that it automatically downloads any V+ updates (and Valheim dedicated Server updates) at runtime. As long as they don't do any breaking changes the update is picked up within 15min after release and automatically applied to your running server.

Hydeo:
All of that might highlight #95 . 
I got my server restarted 4 times today without warning with multiple V+ version changes forcing players to update their client accordingly (with no way for them to know what was the version the server was on).
It was just testing but I can see how this could be a issue, using V+ auto-update means we are at the mercy of their repo. 

I also have some concern about about the .cfg and auto-update, a major update might break breaking changes to this file and that would down the server indefinitely. 

What do you think ?

oeed:
Just wanted to say thanks @Hydeo and @lloesche for your work in getting V+ added in – was excitedly waiting for its merge!

PR #128, opened 2021-02-28
Feature backup to sftp

hasechris:
Hi,

i wanted to copy the backup files onto my local sftp/ssh linux server. I added the needed ENV Variables and the needed rsync function in the valheim-backup script.

Also i added the possible ENV Variables into the ENV Table in the readme file. 

Greetings
hasechris


Dockerfile
```diff
@@ -46,6 +46,7 @@ RUN dpkg --add-architecture i386 \
         jq \
         python3-minimal \
         python3-pkg-resources \
+	ssh \
```
lloesche, line 49: indentation is off

README.md
```diff
@@ -64,6 +64,12 @@ For more deployment options see the [Deployment section](#deployment).
 | `BACKUPS_CRON` | `0 * * * *` | [Cron schedule](https://en.wikipedia.org/wiki/Cron#Overview) for world backups (disabled if set to an empty string or if the legacy `BACKUPS_INTERVAL` is set) |
 | `BACKUPS_DIRECTORY` | `/config/backups` | Path to the backups directory |
 | `BACKUPS_MAX_AGE` | `3` | Age in days after which old backups are flushed |
+| `BACKUPS_SFTP_ENABLE` | `0 (false)` | Enable/Disable SFTP Remote Backup |
```
lloesche, line 67: We're using `true/false` for everything.

defaults
```diff
@@ -35,6 +35,15 @@ BACKUPS_INTERVAL=${BACKUPS_INTERVAL:-315360000}
 BACKUPS_CRON=${BACKUPS_CRON-0 * * * *}
 BACKUPS_DIRECTORY=${BACKUPS_DIRECTORY:-/config/backups}
 BACKUPS_MAX_AGE=${BACKUPS_MAX_AGE:-3}
+# Enable SFTP Copy of Backup zip
+BACKUPS_SFTP_ENABLE=${BACKUPS_SFTP_ENABLE:-0}
+[ "$BACKUPS_SFTP_ENABLE" = true ] && BACKUPS_SFTP_ENABLE=1
```
lloesche, line 40: I don't understand this logic. Why not just check for `[ "$BACKUPS_SFTP_ENABLE" = true ]` in all the lines below where it currently checks for `[ "$BACKUPS_SFTP_ENABLE" = 1 ]`?

valheim-backup
```diff
@@ -23,6 +23,8 @@ main() {
         while [ $run = true ]; do
             backup
             flush_old
+	    sleep 2
```
lloesche, line 26: What is this 2 second sleep for?

valheim-backup
```diff
@@ -23,6 +23,8 @@ main() {
...
+	    rsync_to_sftp
```
lloesche, line 27: Indentation is off.

valheim-backup
```diff
@@ -49,6 +51,11 @@ backup() {
     chmod "$BACKUPS_FILE_PERMISSIONS" "$backup_file"
 }
 
+rsync_to_sftp() {
```
lloesche, line 54: Maybe I missed it, but where does it check if `BACKUPS_SFTP_ENABLE=true`? Wouldn't it execute `rsync` even with `BACKUPS_SFTP_ENABLE=false`?

lloesche (changes requested):

lloesche:
I wonder if this should be prefixed `BACKUPS_SFTP` when really it's `BACKUPS_RSYNC` and not using `sftp` anywhere.

lloesche:
@hasechris what do you think about making this more generic? Like instead of over time adding backup_sftp, backup_rsync, backup_s3, backup_ftp, backup_gcp, backup_azure, etc. we could have a generic post_backup hook that allows you to do whatever you like with the created backup file. The default could be `:` (noop) and if you set it to anything it'll get the name of the created backup file as first argument.

Dockerfile
hasechris: sorry, what do you mean by this?

valheim-backup
hasechris: waiting on harddrive syncing

valheim-backup
hasechris: *facepalm* big oof, i knew it was to late yesterday. Sorry

hasechris:
Hi @lloesche,

hm, a day later and with your comments i'm starting to think i really rushed this idea to much without actually thinking about it.

I just wanted to get the backup files onto my local NAS. Halfway through i sadly found out that sftp cant determine, which files are already on the remote target and are not changed on the local disk. So i switched back to rsync and forgot to change the variable names.


Regarding your second thought (more generic): Yep, that would be the right way to do it; i just made the change only with sftp/rsync in mind. I think we should not reinvent the wheel, so to speak ;) there should be a readymade Script elsewhere which will meet this backup requirement. I will go hunting...

lloesche:
@hasechris no worries I think the idea is great!

Check out https://github.com/lloesche/valheim-server-docker/pull/134 and see if `lloesche/valheim-server:dev` works for your use case.

There's two relevant hooks for now. `POST_BOOTSTRAP_HOOK` and `POST_BACKUP_HOOK`.
The former could be used to install extra packages on boot like `ssh` (although I now included `openssh-client` in the default image) or `awscli` if you wanted to backup to S3.
`POST_BACKUP_HOOK` takes any command and within it replaces the string `@BACKUP_FILE@` with the full path of the just created zip file. I updated the README with some usage examples.

Would actually be cool if you could review the PR! Let me know if you find anything that needs changing to support your use case.

Dockerfile
lloesche: Like the `ssh` is under-indented. As in all other packages are indented with 8 spaces but the line you added is indented with 6 spaces.

valheim-backup
lloesche: This is not a thing. So first, if you wanted to make the kernel flush its buffers to disk you could call `sync`. That is more reliable than sleeping for two seconds. There is no guarantee that fs buffers have been flushed after 2s.
But secondly this really is not needed at all since you are using user space tools like rsync which use libc functions like `open()` and `read()` which in turn use the corresponding kernel functions (syscalls) like `open()` and `read()`. So even if the zip file is in full or partially in a fs buffer that hasn't been flushed to disk yet there is no way for you to not receive the correct data when accessing the file using standard user space tools.
Only if you were to access the raw block device directly, bypassing the kernel's interfaces for accessing the filesystem, would it make any sense to sync buffers to disk. I.e. never.

PR #147, opened 2021-03-03
Added healthcheck

MattiasFestin:
Hi.

I added healthcheck on the UDP ports via netcat. 

I had some problem with unresponsive server, and this solved the issue when the Valheim server does not respond. 
Docker will auto restart the container until it is healthy again.


lloesche:
Hi @MattiasFestin thank you for your PR. However this PR seems untested.

Issues
- `netcat` is being installed inside the build container instead of the running container so the HEALTHCHECK will fail from the start.
- Healthcheck fails if port 2458 is not bound but Valheim currently only uses ports 2456 and 2457 and 2458 is unused. So again the container would be restarted.
- If we wanted to know if a process is listening on those UDP ports we could read the information from /proc/net without the need to install and run netcat.
- If the user adjusts the server port via `SERVER_PORT` env var (many users run multiple instances of this container on multiple ports) the healthcheck would again fail.

I have no idea how this PR could have ever worked on your local system but I suspect it was never tested.

Regardless, the server is being supervised by `supervised`. It is also updated and restarted by processes inside the container. So there are situations where ports 2456 and 2457 are unavailable and that is completely expected during updates. In those situations we wouldn't want Docker to restart the container. So we won't be integrating this into the container.

MattiasFestin:
Ok, it was a bit more complicated then I thought.

I am running the image on my server. I don't know why it passes the health check. 😕 
```
CONTAINER ID   IMAGE                                        COMMAND                  CREATED        STATUS                    PORTS                              NAMES
8be9ef7a495a   mattiasfestin/valheim-server-docker:latest   "/usr/bin/supervisor…"   22 hours ago   Up 22 hours (healthy)   0.0.0.0:2456-2458->2456-2458/udp   valheim-server
```

Will be switching back then.
Thank you for your detailed answer. 👍

schildbach:
Just out of curiosity: has a sensible HEALTHCHECK been developed since closing this issue?

lloesche:
@schildbach you might be interested in https://github.com/lloesche/valheim-server-docker#status-web-server
Also the file `/var/run/valheim-server.status` can be introspected. If it says `running` it means that the server is up and accepting connections on `$SERVER_PORT`.

PR #149, opened 2021-03-03
Adding query port

dhumphrey7:
Adding the option to specify query port to allow multiple server to run under a single IP.


lloesche:
Can you link to some documentation describing the use of the `-queryport` valheim server arg and its exact effects?
Also should this be a dedicated arg or should we introduce something like we did for `steamcmd` where you can define `STEAMCMD_ARGS` to be used for arbitrary arguments. Like a generic `SERVER_ARGS` maybe?

lloesche:
I just checked a running server and only udp 2456 and 2457 are bound by the valheim server process. All other connections are outgoing. What is the exact effect of the `-queryport` arg?

dhumphrey7:
From my understanding the query port is used for steam to check for the server. This appears to be needed when running multiple servers on the same IP to tell steam which port to query for the game.  I know with the default setup I am only able to join game by IP and it is not listed in the server list, not sure if this is related.  Unfortunately, I am not an expert and could possibly be off target here and I apologize if that is the case as I can't find any information that directly relates to Valheim itself.
https://support.steampowered.com/kb_article.php?ref=8571-GLVN-8711
https://answers.unrealengine.com/questions/787258/view.html

lloesche:
I am still not clear on how this relates to Valheim server. The default Valheim server query port is `2457` or rather +1 from `$SERVER_PORT`. But regardless you can now use `SERVER_ARGS="-queryport 123456"` to supply additional arguments to the Valheim server.

Issue #164, opened 2021-03-05
Shutdown Hooks


TheCatLady:
Really love the event hooks and the degree of customizability they provide!  Could we also get hooks for shutdown events?

~~Would also be nice to have hooks that are only triggered if an update is actually available and being applied, rather than on every update check.~~ (Appears that the pre/post restart hooks serve this purpose!)

lloesche:
Implemented in #167

PR #167, opened 2021-03-05
Add pre/post shutdown hooks

TheCatLady:
Will test this in a bit, but since the code is so organized and easy to make sense of, I went ahead and just made the change required for my feature request #164. 🙂

Documentation would also need to be updated, will add that change here shortly.


lloesche:
@TheCatLady thank you for this PR! Makes a lot of sense to have shutdown hooks.
For completeness please add the two hooks to the end of the `defaults` file as well. Otherwise looks good. 🚢

TheCatLady:
> @TheCatLady thank you for this PR! Makes a lot of sense to have shutdown hooks.
> For completeness please add the two hooks to the end of the `defaults` file as well. Otherwise looks good. 🚢

Seems like the `info "Shutdown complete"` logging is actually unreachable since the process is killed by then?

So I removed that hook and logging line.  You're definitely a lot more knowledgeable about this though, so if you have any guidance on getting the post shutdown hook to work, I'd love to learn :)

Will add to the `defaults` file. 👍🏻

lloesche:
That code is part of the `shutdown()` signal handler and will be reached independent of the `exit 0` that follows the server process exiting. So it is fine to have the post_shutdown_hook down there.

See
```
03/05/2021 13:42:21: Net scene destroyed
Binary file (standard input) matches
2021-03-05 13:42:22,556 INFO reaped unknown pid 169
2021-03-05 13:42:22,556 INFO reaped unknown pid 168
2021-03-05 13:42:22,835 INFO waiting for valheim-server to stop
2021-03-05 13:42:24,835 INFO waiting for valheim-server to stop
2021-03-05 13:42:24 [159] (INFO) Shutdown complete
```

lloesche:
FYI I just merged some updates to the README you might have to rebase on latest `main` branch.

TheCatLady:
> That code is part of the `shutdown()` signal handler and will be reached independent of the `exit 0` that follows the server process exiting. So it is fine to have the post_shutdown_hook down there.
> 
> See
> 
> ```
> 03/05/2021 13:42:21: Net scene destroyed
> Binary file (standard input) matches
> 2021-03-05 13:42:22,556 INFO reaped unknown pid 169
> 2021-03-05 13:42:22,556 INFO reaped unknown pid 168
> 2021-03-05 13:42:22,835 INFO waiting for valheim-server to stop
> 2021-03-05 13:42:24,835 INFO waiting for valheim-server to stop
> 2021-03-05 13:42:24 [159] (INFO) Shutdown complete
> ```

Oh hm, I couldn't get the post shutdown hook to trigger.  I'll add it back though.

TheCatLady:
> FYI I just merged some updates to the README you might have to rebase on latest `main` branch.

Already done 😉

lloesche:
> Oh hm, I couldn't get the post shutdown hook to trigger. I'll add it back though.

Hmm maybe there is a race between the two exits and I was just lucky it could get the info printf out in time?
If that is the case we might have to remove the `exit 0` in `run_server()`.

Although I would think that the `sleep` in the shutdown handler takes an eternity longer than the waitpid in `run_server()` (from a computers point of view). Like, what I mean is, as soon as the valheim server process exits the `wait $valheim_server_pid` will be released and `exit 0` in `run_server()` reached. That all happens within few milli- or even microseconds. Whereas the shutdown handler is waiting for the server pid to be gone with a `sleep 6`. So I feel if there was a race I would have had to be incredibly lucky to reach that `info` line in `shutdown()` handler before reaching `exit 0` in `run_server()`

Will have to debug this further. Generally speaking the `shutdown()` signal handler is already doing a lot of things. Usually you'd only want to set a flag and do minimal work in an interrupt. Maybe we'll have to move all of the shutdown code into a subshell and change the signal handler code to something simple like changing a state flag `run=false` or sth. and long those lines.

README.md
```diff
@@ -126,6 +126,8 @@ The following environment variables can be populated to run commands whenever sp
 | `POST_RESTART_HOOK` |  | Command to be executed after a server restart was performed. Future restarts and update checks are blocked until this command returns. |
 | `PRE_SERVER_RUN_HOOK` |  | Command to be executed before the server is started. Server startup is blocked until this command returns. |
 | `POST_SERVER_RUN_HOOK` |  | Command to be executed after the server has finished running. Server shutdown is blocked until this command returns or a shutdown timeout is triggered after 29 seconds. |
+| `PRE_SERVER_SHUTDOWN_HOOK` |  | Command to be executed before the server is shut down. Server shutdown is blocked until this command returns. |
```
lloesche, line 129: We should also make a note that after 90 seconds the entire process will be hard killed by supervisord.
See https://github.com/lloesche/valheim-server-docker/blob/286945daa4f4c98608f9f5357d8792d8be4eaf38/supervisord.conf#L72

lloesche (changes requested):
LGTM would just add the info that if `PRE_SERVER_SHUTDOWN_HOOK` holds the shutdown process for more than 90 seconds the higher instance `supervised` will hard kill the `valheim-server` process and all child processes. This is to ensure we don't ever get stuck on a misbehaving server.
We can increase this 90s timeout if you feel it makes sense. I'm neither here nor there.

TheCatLady:
> LGTM would just add the info that if `PRE_SERVER_SHUTDOWN_HOOK` holds the shutdown process for more than 90 seconds the higher instance `supervisord` will hard kill the `valheim-server` process and all child processes. This is to ensure we don't ever get stuck on a misbehaving server.
> We can increase this 90s timeout if you feel it makes sense. I'm neither here nor there.

I personally don't need the timeout increased, as I'm just issuing a `curl` to a Discord webhook endpoint.

TheCatLady:
> > Oh hm, I couldn't get the post shutdown hook to trigger. I'll add it back though.
> 
> Hmm maybe there is a race between the two exits and I was just lucky it could get the info printf out in time?
> If that is the case we might have to remove the `exit 0` in `run_server()`.
> 
> Although I would think that the `sleep` in the shutdown handler takes an eternity longer than the waitpid in `run_server()` (from a computers point of view). Like, what I mean is, as soon as the valheim server process exits the `wait $valheim_server_pid` will be released and `exit 0` in `run_server()` reached. That all happens within few milli- or even microseconds. Whereas the shutdown handler is waiting for the server pid to be gone with a `sleep 6`. So I feel if there was a race I would have had to be incredibly lucky to reach that `info` line in `shutdown()` handler before reaching `exit 0` in `run_server()`
> 
> Will have to debug this further. Generally speaking the `shutdown()` signal handler is already doing a lot of things. Usually you'd only want to set a flag and do minimal work in an interrupt. Maybe we'll have to move all of the shutdown code into a subshell and change the signal handler code to something simple like changing a state flag `run=false` or sth. and long those lines.

Perhaps we can increase the `sleep 6`?  I'll do some testing 🤔

lloesche:
I just tested the following:
```
...
    info "Shutdown complete"
    dummy_hook
    exit
}

dummy_hook() {
    echo "Sleeping for a really long time"
    dummy_cmd="sleep 10"
    eval "$dummy_cmd"
    echo "Slept a reaaaaly long time"
}
```

then ran `supervisorctl stop valheim-server` inside the container and it worked just fine:
```
03/05/2021 14:54:26: Net scene destroyed
2021-03-05 14:54:27,513 INFO reaped unknown pid 101
2021-03-05 14:54:27,514 INFO reaped unknown pid 100
2021-03-05 14:54:27,900 INFO waiting for valheim-server to stop
2021-03-05 14:54:29,900 INFO waiting for valheim-server to stop
2021-03-05 14:54:29 [94] (INFO) Shutdown complete
Sleeping for a really long time
2021-03-05 14:54:31,900 INFO waiting for valheim-server to stop
2021-03-05 14:54:33,900 INFO waiting for valheim-server to stop
2021-03-05 14:54:35,900 INFO waiting for valheim-server to stop
2021-03-05 14:54:37,900 INFO waiting for valheim-server to stop
2021-03-05 14:54:39,900 INFO waiting for valheim-server to stop
Slept a reaaaaly long time
2021-03-05 14:54:39,952 INFO stopped: valheim-server (exit status 0)
```

So I think the `POST_SERVER_SHUTDOWN_HOOK` you wrote should be good.

TheCatLady:
@lloesche I still couldn't get it to trigger with what I had previously, but I switched the order so that the hook is executed before the logging, and now it works! 👀

lloesche (approved):
> and now it works! 👀

Good enough for me 🚢 😄

TheCatLady:
@lloesche Just FYI, I figured out why my shutdown hook was not triggering consistently... it was Docker not allowing the container enough time to exit gracefully.  I configured `stop_grace_period` in my Compose YAML and all is good now!

lloesche:
@TheCatLady good suggestion! I've added it to the default compose file and also added a `--stop-timeout 120` to all `docker run` commands/documentation.

PR #173, opened 2021-03-05
Support for supervisor http server

lloesche:


TheCatLady:
Nice, this looks great!

P.S.- You'll notice that the Docker Hub CI test is much slower than the GitHub workflow CI 😉  It's not really necessary to have both, and the GitHub one also has the advantage of others being able to view the status / any errors (useful when people other than you submit PRs!).

lloesche:
> You'll notice that the Docker Hub CI test is much slower than the GitHub workflow CI
🙂 I saw that and already swapped them around. Now only GH CI is mandatory.

lloesche:
@TheCatLady Is there a way to build PR's as tagged Docker images, like `:pull173` so they are easy to try out but make it so they get deleted once the PR is closed and/or merged?

TheCatLady:
> @TheCatLady Is there a way to build PR's as tagged Docker images, like `:pull173` so they are easy to try out but make it so they get deleted once the PR is closed and/or merged?

I'm not sure actually.  I didn't set up the workflow to actually push images for pull requests (it'll build but not publish the images).

PR #176, opened 2021-03-06
Move image from stable to stable-slim and add CAP_SYS_NICE for Steam networking library

lloesche:
This PR
- moves the image from stable to stable-slim as suggested by @TheCatLady 
- symlinks many busybox applets that might be useful in PRE/POST hooks
- adds CAP_SYS_NICE for the Steam networking library

The last one gets rid of
```
Warning: failed to set thread priority
```
and might help on very loaded systems with https://github.com/lloesche/valheim-server-docker/discussions/104

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stable as build-env
+FROM debian:stable-slim as build-env
 ARG TESTS
 RUN apt-get update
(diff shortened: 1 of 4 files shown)


TheCatLady:
Ooh nice!  Sorry I didn't get to this last night, I had a beer for the first time in forever and I totally just passed out on the couch lol.  Looks good!

PR #179, opened 2021-03-06
build: Docker optimizations

TheCatLady:
This PR makes some small optimizations to our Docker image and reduces the image size by ~10 MB.  Further optimization is possible but I just wanted to take care of some of the easy things in this PR :)

- set `ENV DEBIAN_FRONTEND=noninteractive` (has the advantage of persisting in the Docker image after build)
- add `--no-install-recommends` in the second build stage to minimize the number of unnecessary packages installed
- ~~add `localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8` to fix the locale gen error~~
- use `curl` to fetch `steamcmd_linux.tar.gz` instead of `COPY`, since `COPY` creates another layer in the final image (when `/tmp` is deleted later, it can't remove stuff from other layers, so currently this tarball is actually present in the final image, albeit hidden)
- consolidate `COPY` commands to reduce number of layers in built image
- remove `VOLUME ["/config", "/opt/valheim"]`, as it is not necessary and only creates an anonymous volume if a volume for `/config` is not defined on container creation
- remove `ENV TZ=Etc/UTC`, since UTC is the default time zone in Docker containers anyway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,6 @@ RUN dpkg --add-architecture i386 \
         jq \
         python3-minimal \
         python3-pkg-resources \
-    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
     && echo 'LANG="en_US.UTF-8"' > /etc/default/locale \
     && echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen \
     && ln -s /usr/local/bin/busybox /usr/local/sbin/syslogd \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM debian:stable-slim as build-env
+ENV DEBIAN_FRONTEND=noninteractive
 ARG TESTS
 RUN apt-get update
-RUN DEBIAN_FRONTEND=noninteractive apt-get -y install apt-utils
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential curl python3 python3-pip shellcheck
+RUN apt-get -y install apt-utils
+RUN apt-get -y install build-essential curl python3 python3-pip shellcheck
 WORKDIR /build/busybox
 RUN curl -L -o /tmp/busybox.tar.bz2 https://busybox.net/downloads/busybox-1.32.1.tar.bz2
 RUN tar xjvf /tmp/busybox.tar.bz2 --strip-components=1 -C /build/busybox
@@ -20,6 +21,7 @@ COPY contrib/* /usr/local/share/valheim/contrib/
 RUN if [ "${TESTS:-true}" = true ]; then shellcheck -a -x -s bash -e SC2034 /usr/local/sbin/bootstrap /usr/local/bin/valheim-* /usr/local/share/valheim/contrib/*.sh; fi
 
 FROM debian:stable-slim
+ENV DEBIAN_FRONTEND=noninteractive
 COPY --from=build-env /build/busybox/_install/bin/busybox /usr/local/bin/busybox
 COPY --from=build-env /build/vpenvconf/dist/vpenvconf-*.linux-x86_64.tar.gz /tmp/vpenvconf.tar.gz
 COPY bootstrap /usr/local/sbin/
@@ -27,12 +29,11 @@ COPY valheim-* /usr/local/bin/
 COPY defaults /usr/local/etc/valheim/
 COPY common /usr/local/etc/valheim/
 COPY contrib/* /usr/local/share/valheim/contrib/
-ADD https://steamcdn-a.akamaihd.net/client/installer/steamcmd_linux.tar.gz /tmp/
 RUN dpkg --add-architecture i386 \
     && apt-get update \
-    && DEBIAN_FRONTEND=noninteractive apt-get -y install apt-utils \
-    && DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade \
-    && DEBIAN_FRONTEND=noninteractive apt-get -y install \
+    && apt-get -y install --no-install-recommends apt-utils \
+    && apt-get -y dist-upgrade \
+    && apt-get -y install --no-install-recommends \
         libc6-dev \
         lib32gcc1 \
         libsdl2-2.0-0 \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -97,8 +100,6 @@ RUN dpkg --add-architecture i386 \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 COPY supervisord.conf /etc/supervisor/supervisord.conf
 
-ENV TZ=Etc/UTC
```
lloesche, line 100:
```
lukas@MACROMI:~$ docker run --rm -it debian:stable-slim /bin/bash
root@390ed9c45055:/# echo $TZ

root@390ed9c45055:/#
```

lloesche: Without this supervisor can not query the TZ env var unless the user has explicitly specified it.
https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

All the scripts are fine as they will get it from `defaults` but supervisor isn't.

lloesche: Actually, now that we have the new `bootstrap` script and supervisord is no longer the entrypoint we could get it from `defaults` before the supervisor `exec`. I'm fine with either.

lloesche (changes requested):
All looks good except for the `TZ` thing. Without it supervisord and esp. `crond` will use whatever is the libc's default. Which on Debian might be UTC but I like to be explicit.

Dockerfile
TheCatLady: @lloesche Yeah I think it's best to not define that here in the Dockerfile.  How would we add it to the `bootstrap` script?  Would `export TZ="$TZ"` before the `exec` work?  (This is a bit out of my wheelhouse 😛)

lloesche: Just `export TZ` right before the exec is enough since the value is already imported on line 5.

TheCatLady:
Also just want to make a note here that we still currently have `vpenvconf.tar.gz` in the final built image, even though it is hidden since it was `rm`'d in the layer created by the `RUN` command.  I wasn't sure about the best way to handle that so I didn't touch it in this PR.

TheCatLady:
> Also just want to make a note here that we still currently have `vpenvconf.tar.gz` in the final built image, even though it is hidden since it was `rm`'d in the layer created by the `RUN` command (similar to `steamcmd_linux.tar.gz` before). I wasn't sure about the best way to handle that so I didn't touch it in this PR.

Actually... let me try to fix that in this PR.  I have an idea that would further reduce the number of image layers in our final built image.

lloesche:
Do we have a list of the packages we no longer have through `--no-install-recommends`? I just wanna make sure there was nothing that we implicitly used because I might have thought it is part of the base image when really it came through a recommends.

Also would be handy if I could `docker pull http://ghcr.io/lloesche/valheim-server:PR179` to test these changes before merging. From reading the changes I'm pretty confident that they are fine, but on my local system I can usually run a few test containers, connect with a client, etc.

TheCatLady:
> Do we have a list of the packages we no longer have through `--no-install-recommends`? I just wanna make sure there was nothing that we implicitly used because I might have thought it is part of the base image when really it came through a recommends.

I don't, but I did build and deploy my changes locally, and hubby and his friends haven't complained that I broke anything yet lol.  The only thing I can think of is that we don't use Valheim Plus, so I wasn't able to test that.

> Also would be handy if I could `docker pull http://ghcr.io/lloesche/valheim-server:PR179` to test these changes before merging. From reading the changes I'm pretty confident that they are fine, but on my local system I can usually run a few test containers, connect with a client, etc.

The best way to do that sort of testing is to pull the changes and build it locally.  GHCR doesn't currently have a way to allow automated deletion of images, so it's not advisable to build and push on every PR commit.

lloesche:
> The best way to do that sort of testing is to pull the changes and build it locally

yea but, you know... why do what a computer can do for me 🙂

TheCatLady:
> yea but, you know... why do what a computer can do for me 🙂

If you'd like, a better option is to publish PR images under a different name (like `lloesche/valheim-server-test`), and leave that package as private.  We can definitely tag images with the PR number, I just don't think it's wise to have those tags published under the main package since it'd be visible to everyone and cleanup of images from merged/closed PRs would need to be done manually.

lloesche:
> Also just want to make a note here that we still currently have `vpenvconf.tar.gz` in the final built image

That's like few kb or so. Nothing to lose sleep over.

Dockerfile
```diff
@@ -52,6 +53,7 @@ RUN dpkg --add-architecture i386 \
         jq \
         python3-minimal \
         python3-pkg-resources \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
```
lloesche, line 56: What error are we talking about? For me locale-gen just produces:
```
Generating locales (this might take a while)...
  en_US.UTF-8... done
Generation complete.
```
and that's it?

TheCatLady: It was failing when I built the image locally.  I'll switch my local repo back to the `main` branch and build to locate the exact error message once I finish testing my latest changes.

lloesche:
> cleanup of images from merged/closed PRs would need to be done manually

So there's no way to have an action run whenever a PR is merged and have the number of the just merged PR handed to that action? I mean that's all the information required to delete the image that was created for that PR, right?
I could just build something for that and run it on Travis but I can't imagine Github can't do that. But then I have zero experience with GH Actions.

TheCatLady:
> So there's no way to have an action run whenever a PR is merged and have the number of the just merged PR handed to that action? I mean that's all the information required to delete the image that was created for that PR, right?
> I could just build something for that and run it on Travis but I can't imagine Github can't do that. But then I have zero experience with GH Actions.

Unfortunately GHCR is still relatively new, and doesn't support that yet.  The only way to delete GHCR images is to do so manually AFAIK.

Dockerfile
lloesche: Ok let me know. Because all `locale-gen` is doing is calling `localedef` so we only need the one or the other. But as far as I can tell it's already good. There used to be a locale error in a very very old version, like three weeks back or so.

lloesche: Oh could it be that you executed commands before `locale-gen`? That's when you would see errors because the locale doesn't exist yet.

TheCatLady: Huh I can't seem to repro this anymore (was seeing it consistently earlier somehow??).  I'll remove this change for now and just keep an eye out to see if it happens again.

lloesche (approved):
Thank you, LGTM 🚢 👍

Issue #184, opened 2021-03-06
Question - (DEBUG) Valheim Server is not yet downloaded - waiting


chrsolr:
I use Docker for Mac and I keep getting "(DEBUG) Valheim Server is not yet downloaded - waiting" message on the logs.

`docker-compose.yaml`:
```
version: "3"

services: 
  valheim: 
    image: lloesche/valheim-server
    cap_add:
      - sys_nice
    volumes: 
      - $HOME/Documents/valheim/valheim-server/config:/config
      - $HOME/Documents/valheim/valheim-server/data:/opt/valheim
    ports: 
      - "2456-2458:2456-2458/udp"
      - "9001:9001/tcp"
    env_file:
      - $HOME/Documents/valheim/valheim-server/valheim.env
    restart: always
```
and `valheim.env`:
```
SERVER_NAME="Z8OTester"
WORLD_NAME=WORLD
SERVER_PASS=secret
SERVER_PUBLIC=true
```

lloesche:
Can you paste the entire log? Without it I can just guess.
Could it be one of OSX' new security features where not every process is allowed to read and write every directory on the system?

chrsolr:
```
Creating network "valheim-server_default" with the default driver
Creating valheim-server_valheim_1 ... done
Attaching to valheim-server_valheim_1
valheim_1  | 2021-03-06 22:48:25,319 WARN No file matches via include "/etc/supervisor/conf.d/*.conf"
valheim_1  | 2021-03-06 22:48:25,320 INFO Set uid to user 0 succeeded
valheim_1  | 2021-03-06 22:48:25,335 INFO RPC interface 'supervisor' initialized
valheim_1  | 2021-03-06 22:48:25,336 INFO supervisord started with pid 1
valheim_1  | 2021-03-06 22:48:26,341 INFO spawned: 'valheim-bootstrap' with pid 13
valheim_1  | 2021-03-06 22:48:26 [13] (INFO) Setting timezone Etc/UTC
valheim_1  | 2021-03-06 22:48:26,373 INFO success: valheim-bootstrap entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
valheim_1  | 2021-03-06 22:48:26 [13] (DEBUG) Creating cron to do world backups using schedule 0 * * * *
valheim_1  | 2021-03-06 22:48:26 [13] (DEBUG) Creating cron to check for updates using schedule */15 * * * *
[...]

Issue #170, opened 2021-03-05
Healthchecks options for monitoring


brandtkeller:
Less of a request and more of general knowledge gathering (if available).

I am writing my own orchestration for this image on kubernetes following some design standards that I feel are important/different from other implementations.

As such, I was curious if there are any exposed options for monitoring the health of the server?

These may be actions that establish - "Has the game server fully started?", "Is the game server still running?"

Any help would be appreciated!
Thanks

sisu4u:
Not sure if that's possible easily. k8s does not support udp health checks: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#tcp-probes.
An alternative would be some sort of script inside the container that checks the server connection when it's called and return exit code != 0 iff failed. K8s can then use this script as a health check.

lloesche:
@brandtkeller I just added something for you in #173 that might be helpful. The process supervisor I'm using inside the container provides an XML-RPC API. This API can provide status information about the containers different services. The most interesting one being `valheim-server` I suppose. Details here https://github.com/lloesche/valheim-server-docker#supervisor

I know, XML-RPC, likely not something you can plug natively into a livenessProbe but I figured baby steps. We'll see if it is useful for this issue and if not we can always look for other options.

A simpler readinessProbe could also be to just check if `/var/run/valheim-server.pid` exists. If that's the case the valheim_server is running. That does not mean however it is currently accepting connections. It could be in the process of starting up. It could also be constantly crashing and in a restart loop.

Then for pretty high reliability that Valheim server is running and ready to accept connections there is `/proc/net/udp6` and `/proc/net/udp`.

You could check if anything is bound on `$SERVER_PORT`. To find out you would take the second field split at the colon and convert the second field of this split from hex to dec and compare to `$SERVER_PORT`.
A one-liner could look like this
```
awk -v search_port=2456 'BEGIN {ec=1} {if ($1 ~ /^[0-9]/) {split($2, local_bind, ":"); port=sprintf("%d", "0x" local_bind[2]); if (port == search_port){ec=0}}} END {exit ec}' /proc/net/udp*
```
If this exits `0` a process is listening on `search_port` and if it exits `1` no process was found that bound to this port.

lloesche:
Thinking about it, I could just integrate that last check into the `valheim-server` start script and produce a more easily consumable status file for you.

[...]

Issue #158, opened 2021-03-04
How should Valheim Plus config updates be handled?


kachunkachunk:
[edit: This is not so much an issue/bug, but a question and suggestion]

When Valheim Plus updates, it often comes with configuration changes, and a new config file. Still, I've been wondering - how does the update checker service handle the new config file that's included with each updated V+ release that's obtained and installed?

I assume that it doesn't overwrite the config file that's already generated or used when the container is instantiated, so as V+ updates and the server restarts, the old configuration file continues to load. Thus you don't use the new sections or key=value properties, post-restart.
If my assumption is correct, that's probably fine (and sane), since V+ will assume defaults for any undefined config file properties.
If that's incorrect, and the **new defaults** are loaded instead, it may pose issues for users and I'd have to suggest avoiding that.
For instance, maybe the max item stack size decreased from the user value to defaults, and thus all their item stacks will truncate.

I've otherwise been thinking that the environment variable config generation is a pretty smart way of making sure some things are persistent across versions (assuming key=value pairs don't change drastically). At least on container creation.
Maybe I can suggest that you take the new V+ config file (included with each update and populated with defaults), read in the container environment variables or old config file, and merge the key=value pairs into the new config?
It's easier said than done, though; I have not yet figured out a good process to get one config's set into another config (without depending on sorting, expecting the same sections or line numbers, etc). I've tried and did not really come up with a simple solution to this with shell scripting. : /

But, in case it helps, I have written [this](https://github.com/kachunkachunk/valheim-linux-server/blob/container-support/scripts/generate_valheim_plus_env.sh
) shell script that will read in a config file (default or customized) and it into a list of environment variables that you can load with `docker run`. You're free to do whatever you want with it, if you find it or any part of it helpful (note that the link/branch _may_ be invalid in some days/weeks). I'm unsure yet if I need to rely on environment variables, or config files, so would be nice to know what your thoughts are on everything concerning updates vs configs.

lloesche:
@kachunkachunk great question! Here is how it currently works:

When downloading/updating V+ if the config folder `/config/valheimplus/` does not exist it is created and default configs (`valheim_plus.cfg` and `BepInEx.cfg`) copied from the ZIP archive. If the folder exists files in it are not touched.

If a user defines sections using the `VPCFG_` prefixed env variables then only those are written into the config file. However if an existing config file is found it is read in first and then the env config values are merged on top of the existing config. A new config file is written and the old one moved to the `.old` extension.

[...]

PR #185, opened 2021-03-06
Implement server status file; Copy V+ default config

lloesche:
Adds an undocumented env variable SERVER_STATUS_FILE which defaults to /var/run/valheim-server.status

The file contains one of the following
```
bootstrapping
starting
running
stopping
stopped
```

Whenever the status is `running` it means that valheim-server is listening on the UDP port configured as `SERVER_PORT`.
This is to help with #170 

PR also
- copies the default ValheimPlus config from new ZIP archives to /config/valheimplus/valheim_plus.cfg to help with #158 
- adds extra checks to detect if Valheim or ValheimPlus download failed #184
- adds an undocumented debug env var `DEBUG_START_FRESH` which if `true` wipes the data directories


Dockerfile
```diff
@@ -35,11 +35,12 @@ COPY contrib/* ./share/valheim/contrib/
 FROM debian:stable-slim
 ENV DEBIAN_FRONTEND=noninteractive
 COPY --from=build-env /artifacts/ /usr/local/
+COPY supervisord.conf /etc/supervisor/supervisord.conf.valheim
```
TheCatLady, line 38: Instead of copying this file on its own, we should copy it during the first build stage into `/artifacts`, and then `mv` and `chmod` it in the `RUN` command below during this second stage to avoid creating an extra unnecessary layer.  The file itself may be small, but Docker image layers aren't free and we should try to minimize the number of layers in the final image 🙂  Not really a huge deal, but just best practice.

lloesche: Great idea! I also want to optimize the redundant COPYs in the build-env but need to fix the paths first. But I'm going to sleep now. 5:22 am here. Tomorrow! 🙂

TheCatLady: @lloesche The `build-env` isn't as important since it basically gets thrown away, but yeah it'd be nice to just `COPY` once there!

lloesche: Yup I didn't mean from a Docker point of view but from a maintainability one. It's error prone to have the scripts in there twice. Whenever we add or remove one we have to remember to do it in both places.

PR #193, opened 2021-03-08
Refactor Dockerfile and status updater

lloesche:


TheCatLady:
This change increased the image size by 10 MB again. ☹️  Copying `/usr/local` from the build stage doesn't work, since we don't want to copy anything else that may be in there.

lloesche:
There is nothing in there. It is an empty directory structure by default. Dunno where the 10 MB are coming from since this even avoids the tarballs that are otherwise copied from /usr/local/tmp, but I'll take readable non-redundant code over 10 MB image size every day.

```
lukas@SHOEBOX:~/repo/valheim-server-docker$ docker run --rm -it debian:stable-slim find /usr/local/
/usr/local/
/usr/local/src
/usr/local/include
/usr/local/man
/usr/local/games
/usr/local/etc
/usr/local/lib
/usr/local/share
/usr/local/share/man
/usr/local/sbin
/usr/local/bin
```

TheCatLady:
@lloesche Let me look into what specifically caused the image size to increase.  It's odd.

Dockerfile
```diff
@@ -22,25 +26,27 @@ COPY valheim-* /usr/local/bin/
...
+            /usr/local/bin/valheim-updater \
+            /usr/local/bin/valheim-plus-updater \
+            /usr/local/share/valheim/contrib/*.sh \
+        ; \
+    fi
+RUN mkdir -p /usr/local/tmp
+WORKDIR /
+RUN mv /build/busybox/_install/bin/busybox /usr/local/bin/busybox
+RUN tar xzvf /build/vpenvconf/dist/vpenvconf-*.linux-x86_64.tar.gz
+RUN tar xzvf /build/python-a2s/dist/python-a2s-*.linux-x86_64.tar.gz
```
TheCatLady, line 44: What is the purpose of these two lines?  Nothing resulting from this ever gets copied to the second build stage.  I think it's best to copy the tarballs into `/usr/local` (or whichever directory we are copying from in the second stage) and then extract them there.

lloesche: What do you mean? These extract directly to `/usr/local` like they did in the main container. Just that they do it now in the build one. See the build output.

TheCatLady: Ohhhh you're right, missed that.

PR #196, opened 2021-03-08
fix: Docker image size bloat

TheCatLady:
This PR cleans out the `/usr/local` directory before copying files into it, fixing the recent ~10 MB increase in image size.

Also copies the `supervisord.conf` file during the first build stage to avoid creating an extra layer in the final image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,8 @@ RUN python3 setup.py bdist --format=gztar
 WORKDIR /build
 RUN git clone https://github.com/Yepoleb/python-a2s.git \
     && cd python-a2s \
-    && python3 setup.py bdist --format=gztar
+    && python3 setup.py bdist --format=gztar \
+    && rm -rf /usr/local/
 COPY bootstrap /usr/local/sbin/
 COPY valheim-* /usr/local/bin/
 COPY defaults /usr/local/etc/valheim/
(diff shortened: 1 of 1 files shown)


lloesche:
Can you explain where the 10 MB are coming from? /usr/local is empty when you start a fresh `debian:stable-slim` image.

TheCatLady:
> Can you explain where the 10 MB are coming from? /usr/local is empty when you start a fresh `debian:stable-slim` image.

Appears to be 12 MB in `/usr/local/lib/python3.7`, as well as some other smaller files we shouldn't need in the final image.

TheCatLady:
@lloesche Highly recommend [`dive`](https://github.com/wagoodman/dive), btw.  It's great for stuff like this.

lloesche:
Ah found it. The `pip` installs store their data in `/usr/local`. But that means we can't `rm -rf` in the beginning.
We need to `rm -rf /usr/local/lib/python3.7/dist-packages` right before the two `tar` calls.

TheCatLady:
> Ah found it. The `pip` installs store their data in `/usr/local`. But that means we can't `rm -rf` in the beginning.
> We need to `rm -rf /usr/local/lib/python3.7/dist-packages` right before the two `tar` calls.

Ahhh right, will fix that.

lloesche (approved):
👍 Thank you!

PR #202, opened 2021-03-09
add --build flag to docker-compose up command

Shawak:
If you didn't build the image yet you will have to use the `--build` flag.


lloesche:
? The compose file is referencing an image that gets downloaded from Dockerhub. Why would you build anything? And how? If you follow the instructions you don't even have a Dockerfile to build from. I'm confused by this PR.

Shawak:
Nevermind I'm stupid, didn't see it was ment to use the docker-compose file only without cloning the project.

PR #215, opened 2021-03-11
Update README.md

podbiker:
Added changing of the startup while updating command in Portainer


lloesche (approved):
Thank you @podbiker this looks great! I'll download and add the images to this repo but other than that 🚢 👍

PR #241, opened 2021-03-19
use --pull=always

greyltc:
Docker version 19.09 added some new pull logic[1] to the run command that could be used here
[1]: https://github.com/docker/cli/pull/1498


lloesche (approved):
Thank you @greyltc! lgtm 👍 🚢

PR #248, opened 2021-03-20
ci: write commit tag to env var

TheCatLady:
Writes ~~GitHub repository URL, branch, and~~ commit tag/hash to environment variable~~s~~ in the build environment.

Also tags published images with the commit tag/hash.

**Edit:** @lloesche, do we actually need the repository URL or branch name here, or just the commit tag/hash? The method used to obtain the branch name doesn't work for pull requests.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -45,7 +45,11 @@ jobs:
           file: ./Dockerfile
           platforms: linux/amd64
           push: ${{ github.event_name != 'pull_request' }}
-          tags: ghcr.io/lloesche/valheim-server:latest
+          build-args: |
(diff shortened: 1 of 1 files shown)


lloesche:
This is perfect! Will old tags eventually be released though? If not I'd just skip those. Realistically nobody is ever going to manually pull a certain commit so we're just wasting Github's disk space.

lloesche (approved):
👍 🚢

TheCatLady:
> This is perfect! Will old tags eventually be released though? If not I'd just skip those. Realistically nobody is ever going to manually pull a certain commit so we're just wasting Github's disk space.

This does not retroactively apply to previous builds, if that's what you're asking.  We will just tag images with the hash going forward.

PR #252, opened 2021-03-21
build(docker): do not invalidate build cache

TheCatLady:
Write commit hash at the end of the first build stage rather at the beginning, so that caching of the first stage is not invalidated.


lloesche (approved):
Thank you so much for always optimizing my Docker related stuff!! I appreciate it a lot!

PR #255, opened 2021-03-22
Dockerfile: Install uuid package

schildbach:
In order to send Matrix messages via API, a UUID must be supplied. One could use something like `date +%s-%N`, but having the `uuid` command available would be nicer.


lloesche:
There's no need to install extra packages or run date: https://docs.python.org/3/library/uuid.html
```
$ python3 -c 'import uuid; print(uuid.uuid1())'
75898040-8b3b-11eb-882d-00155d981eed
```

Issue #282, opened 2021-03-28
Implementation of BepInExPack


lloesche:
Hello 

Are there any plans to implement BepInExPack similar as the Valheim+ mod?
That would be a great way for easy mod module loading. 

[Mod Link](https://valheim.thunderstore.io/package/denikson/BepInExPack_Valheim/)

_Originally posted by @madhub0815 in https://github.com/lloesche/valheim-server-docker/discussions/279_

lloesche:
@madhub0815 check `lloesche/valheim-server:dev` and specify `BEPINEX=true`. It'll download and install BepInExPack. Right now after adding plugins to the plugins directory whenever there is a Valheim Server update or a BepInExPack update it would get overwritten.
From your point of view, what would be better:
- being able to specify a list of mods to be automatically updated/downloaded and installed from https://thunderstore.io/
- having to manually install and update plugins in a volume mount

I can see use cases for both.

madhub0815:
First i will thank you for the quick implementation. 

To set a list of plugin and get it auto updatet and downloaded would be great.
 
Because some plugins for example equipment and quick slots have to be there when a player is joining the server because if there is a server update and the plugin would not be there the player can not use the equiped amor. 

Therefore the admin have to react manuel and install plugins again wich may be not always possible? 

I will set up a test server later that day to test the general function.

lloesche:
@madhub0815 ok makes sense. Maybe we should also build something that will make all plugins available to the user for download as one ZIP file? So that when the server auto updates like 50 plugins the user doesn't have to do that by themselves?
Do you know if Valheim Plugins are always within the `BepInEx/plugins/` folder? Or are some outside of this directory structure?

madhub0815:
@lloesche Yes plugins should be all in the mentioned folder. The idea of auto update and installing these to clients directly is good but if clients are using a mod manager like "Thunderstore" not practicable. 

Because the folder of the modmanager is "C:\Users\WINDOWSUSER\AppData\Roaming\Thunderstore Mod Manager\DataFolder\Valheim\profiles\MODMANAGERPROFILE\BepInEx\plugins" and plugins have to be managed in the mod manager like turn on/off install or uninstall or just play vanilla without any mods. Because of that you can have mods without changing/editing files in the game directory itself.

Therefore i think it will be easier to give the responsibility of plugins are up to date to the clients themselves by using the mod manager.

[...]

PR #288, opened 2021-03-29
Initial BepInExPack mod support (Implements #282)

lloesche:
This PR unifies BepInEx mod support. It now supports installation and update of ValheimPlus and BepInExPack.

Plugins can be installed in `/config/bepinex/plugins/` for BepInExPack or `/config/valheimplus/plugins/` for ValheimPlus.

Config generation via ENV variables now works for `valheim_plus.cfg` and `BepInEx.cfg` as well as any other `.cfg` files via a `POST_BEPINEX_CONFIG_HOOK` event hook. The config generation tool has been renamed to `env2cfg`.

Please test this PR at `lloesche/valheim-server:dev` and provide feedback below or in #282.


lloesche:
Thanks, fixed!

thetaiter:
Will this add the ability to modify BepInEx.cfg values with env variables similar to what is currently supported for valheim_plus.cfg values?

lloesche:
@thetaiter not sure. BepInEx uses sections with `.` in the name which is not an allowed env var character.
E.g.
```
[Logging.Console]
Enabled = false
```
Not sure how we would encode `Logging.Console`. We could of course come up with arbitrary rules like three underscore mean a dot, two underscore mean an underscore or something like that so that we would set `BEPINEXCFG_Logging___Console_Enable=false`. Not pretty but workable.

lloesche:
@thetaiter the answer is now yes. In https://github.com/lloesche/valheim-server-docker/pull/288/commits/1ec2b5e771b7fea48692a00ecbf9d0c03ddc5920 I reworked the existing config generator and made it more universal. It can now generate/update arbitrary cfg files from env variables. Be it `BepInEx.cfg` or a plugins custom config file. There's also now a `PRE/POST_BEPINEX_CONFIG_HOOK` that allows you to call the config generator whenever BepInEx config is being generated or updated.

thetaiter:
Amazing, thank you! Great work on this project, very impressive.

Issue #299, opened 2021-04-03
BepInEx Mod fails to load with "Failed to extract and install BepInEx - retrying later"


vogeldm:
Attempting to setup a new server. Kept docker config to bare minimum, server fails to load with "Failed to extract and install BepInEx - retrying later"

[_valheim-server_logs.txt](https://github.com/lloesche/valheim-server-docker/files/6253122/_valheim-server_logs.txt)

lloesche:
@vogeldm thank you for your report! From the log it looks like it is failing somewhere in these lines:
https://github.com/lloesche/valheim-server-docker/blob/7e0da99e1d2e77916cffbe5ce9df34a95bc93c37/common#L411-L420

Most likely in `ensure_permissions()`.

I updated the container with more debug log and also made config permission change errors non-critical. Can you pull the latest version of the container and see if it makes a difference? If not please attach a new log with the added debug output.

vogeldm:
Seems like the changes corrected the issue. I've included the logs for reference.
[_valheim-server_logs.txt](https://github.com/lloesche/valheim-server-docker/files/6259676/_valheim-server_logs.txt)

PR #300, opened 2021-04-04
common script file perm check

tacitDev:
I ran into a situation when using Kubernetes NFS volumes and volume mounts in the /config dir due to the `ensure_permissions()` chmod. This workaround allows empty or existing NFS directories that already have the correct permissions to be mounted

### DIFF
--- a/common
+++ b/common
@@ -116,7 +116,6 @@ are_permissions_equal() {
     
     if [ $(stat -c "%a" "$2") -eq $1 ]
     then
-        echo $(stat -c "%a" "$2")
         true  
     else
         false
--- a/common
+++ b/common
@@ -108,16 +108,71 @@ printline() {
 }
 
 
-ensure_permissions() {
-    chmod "$CONFIG_DIRECTORY_PERMISSIONS" /config
-    chmod -f "$CONFIG_FILE_PERMISSIONS" /config/*.txt
-    if [ -d /config/worlds ]; then
-        chmod "$WORLDS_DIRECTORY_PERMISSIONS" /config/worlds
-        chmod "$WORLDS_FILE_PERMISSIONS" /config/worlds/*
+are_permissions_equal() {
+    #local target_umask=$1
+    #local target_path=$2
+    #local found_umask=$(stat -c "%a" "$target_path")
+    #echo $found_umask
+    
+    if [ $(stat -c "%a" "$2") -eq $1 ]
+    then
+        echo $(stat -c "%a" "$2")
+        true  
+    else
+        false
     fi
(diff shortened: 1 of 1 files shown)


lloesche:
Hi @tacitDev thank you for your PR. What is the error this PR is solving?
Also can you check if the latest version of the container already solves your issue? For #299 I changed the way permission change errors are handled and made them non-critical.

tacitDev:
Hi @lloesche! Thanks for the awesome Docker image. After re-reading the PR description, I was a bit vague :laughing:. 

Now that I've stepped back, the problem might be fixable on my end via changing something with my NFS setup (currently using `root_squash,subtree_check,rw`). I'll update if it works, and will post example logs for posterity and searchability in case anyone runs into this issue in the future. For what it's worth, it's similar to this issue here: https://github.com/kubernetes/kubernetes/issues/54601#issuecomment-339928466

For the record, I was using the latest image in Docker Hub as it was pulled more than a few times today while troubleshooting.

The issue is related to using the image via a Kubernetes deployment. I have an NFS service on a different machine that stores config files. When mounting the NFS directory in the pod, with pre-populated files that have correct permissions, `valheim-docker` throws errors when trying to change their permissions (even if they're set to be the same). This caused the `valheimplus` merge step to fail. The container is able to read/write to the file fine, but not change the permissions.  

So, for example:
- NFS dir is `/valheim`, which is mapped to `/config` in the container
  - it has `./worlds` with `.fwl/.db` files and `./valheimplus` with `.cfg` files created 
  - dirs are set to 777 and files set to 666 for testing
- Kubernetes cfg is as follows:
```
[...]
# from spec.template.spec:
      volumes:
      - name: nfs-share-valheim
        nfs: 
          server: <ENDPOINT_HERE>
          path: /valheim
[...] 
# from spec.template.spec.containers::
        env:
        - name: PERMISSIONS_UMASK
          value: "000"
[...]
        volumeMounts:
          - name: nfs-share-valheim
            mountPath: /config
```

With this setup, I was getting logs similar to (might not be word for word, have to recreate it later):
```
chown: changing ownership of /config/worlds/world1.fwl`
Failed to extract and install ValheimPlus
```

The changes in `common` > `ensure_permissions()` of this PR check if the directory and files are already set to the correct permissions before setting them with `chmod`.

Anyway, for now, I'll close this PR.

lloesche:
Thanks for the description.

> throws errors when trying to change their permissions

The latest image of the container (from like 5 min before I originally answered to this PR) should make all permission related errors in `ensure_permissions()` non-critical in the BepInEx/ValheimPlus install process. So you will still see the error in the log file, which I think is valuable info when debugging issues down the line, but it will no longer abort the ValheimPlus/BepInEx install process.

PR #306, opened 2021-04-09
Don't expand cron wildcards

### LINKED ISSUE #304: Server always seems to restart at 1am EST
First, thank you for setting up the nice docker. Working pretty well so far.
I've noticed that the server at 1am EST will restart every day, even when players are logged in.
When I set up the server, I only specified the ports, volumes for config and data, server and world name, and password.

My understanding is that the server should only perform an update by default when no one is logged in.
I've tried figuring out where the server log files should be, to confirm what is happening at 1am, but I'd guess it is always updating at that time?

Ultimately I'd like to figure out what is happening at 1am, and change the time at least to a better time, since I've had players on a few hours after 1am, and it's annoying getting booted in the middle of things.

Thanks!

conorsch:
Disable shell wildcard expansion when defining the crontabs, so that asterisks are literals, not converted into globs of filepaths.

Intended to resolve #304, but could definitely use some more testing. My friends are currently playing on the server I host, so it'll be another 12h or so until I try these changes myself. :wink:

### DIFF
--- a/defaults
+++ b/defaults
@@ -37,14 +37,19 @@ UPDATE_IF_IDLE=${UPDATE_IF_IDLE:-true}
 
 # What time we restart the valheim-server
 # This is usful to mitigate the effects of memory/resource leaks
-RESTART_CRON=${RESTART_CRON-0 5 * * *}
+# Briefly disable wildcard expansion so asterisks are literal
+set +f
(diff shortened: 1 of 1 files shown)


lloesche:
Your assumptions are wrong because you're using the variable wrong. The expansion is happening on your shell because you are not quoting the variable. Not during assignment.

PR #332, opened 2021-04-27
Small V+ documentation changes

GimpArm:
Changed all the V+ examples to have enforceMod=true so anyone copying the examples will immediately have a working setup.
Added a link to the complete V+ configuration documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ Example:
 ```
 -e VALHEIM_PLUS=true \
 -e VPCFG_Server_enabled=true \
--e VPCFG_Server_enforceMod=false \
+-e VPCFG_Server_enforceMod=true \
 -e VPCFG_Server_dataRate=500 \
 -e BEPINEXCFG_Logging_DOT_Console_Enabled=true
 ```
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ Example:
 ```
 -e VALHEIM_PLUS=true \
 -e VPCFG_Server_enabled=true \
--e VPCFG_Server_enforceMod=false \
+-e VPCFG_Server_enforceMod=true \
 -e VPCFG_Server_dataRate=500 \
 -e BEPINEXCFG_Logging_DOT_Console_Enabled=true
 ```
@@ -321,7 +321,7 @@ turns into `/config/valheimplus/valheim_plus.cfg`
 ```
 [Server]
 enabled=true
-enforceMod=false
+enforceMod=true
 dataRate=500
 ```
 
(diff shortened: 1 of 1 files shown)


InB4DevOps:
There are a few changes that do not require EnforceMod to be enabled (eg Password & datarate) so I wouldn't want to have the documentation changed where it's not needed..

README.md
```diff
@@ -312,7 +312,7 @@ Example:
 ```
 -e VALHEIM_PLUS=true \
 -e VPCFG_Server_enabled=true \
--e VPCFG_Server_enforceMod=false \
```
InB4DevOps, line 315: This is wrong. For changing the datarate enforceMod must not be true.

README.md
```diff
@@ -321,7 +321,7 @@ turns into `/config/valheimplus/valheim_plus.cfg`
 ```
 [Server]
 enabled=true
-enforceMod=false
```
InB4DevOps, line 324: see above

README.md
```diff
@@ -640,12 +640,12 @@ To do so enable ValheimPlus (`VALHEIM_PLUS=true`) and configure the following se
...
```
InB4DevOps, line 643: see above

README.md
```diff
@@ -640,12 +640,12 @@ To do so enable ValheimPlus (`VALHEIM_PLUS=true`) and configure the following se
...
+enforceMod=true
 dataRate=600
 ```
 (Or whatever `dataRate` value you require. The value is in kb/s with a default of 60.)
 
-Alternatively start with `-e VPCFG_Server_enabled=true -e VPCFG_Server_enforceMod=false -e VPCFG_Server_dataRate=600`.
```
InB4DevOps, line 648: see above

README.md
```diff
@@ -654,10 +654,10 @@ To do so enable ValheimPlus (`VALHEIM_PLUS=true`), set an empty password (`SERVE
 ```
 [Server]
 enabled=true
-enforceMod=false
+enforceMod=true
 disableServerPassword=true
 ```
-Alternatively start with `-e VPCFG_Server_enabled=true -e VPCFG_Server_enforceMod=false -e VPCFG_Server_disableServerPassword=true`.
```
InB4DevOps, line 660: wrong as well

GimpArm:
If there are so many instances where you don't need/want enforceMod=true, then someone who understands this a lot better than I do needs to describe better what this actually does and when you want to enabled/disable it. Which is why I originally suggested in my ticket that it should be documented better.

README.md
GimpArm: This line is in the Translation table section and has nothing to do with datarate.

README.md
GimpArm: also see above

GimpArm:
V+ also doesn't really document it. It appears exactly one time in their github readme and zero times on their config documentation page. That's who should be adding this. I had assumed this was an undocumented flag you added since you asked me to document it here. When no one actually writes what this flag does in their documentation, how can a user write anything about it? Asked me to write what would have helped me, that would have been an example that actually shows how to turn V+ on and not "just turn some parts on" with no one documenting which parts are turned on and when.

I realize now my issue should have been with V+, thank you for your help.

PR #347, opened 2021-05-16
Specify uid gid

martinfelis:
Feature overview:
- Adds non-privileged user `valheim-server`
- uid and gid of `valheim-server` can be specified as environment variables `USER_ID` and `GROUP_ID` and defaults to 1001.
- permissions for files needed for running valheim are adjusted during container startup by bootstrap process.

Internal changes
- valheim-server, valheim-bootstrap, valheim-backup, and valheim-updater are running with user valheim-server
- use debian's cron instead of busybox (setuid required for valheim-updater to modify crontab, but don't want to use setuid for all busybox binaries)


lloesche (changes requested):
@martinfelis thank you for your PR! Looks super clean, I like it a lot!

That said, any change to the image must be fully backwards compatible to the current image so we don't break existing user's setups.

This PR changes the default UID/GID from 0:0 to 1001:1001. I already know of use cases where this will fail when e.g. trying to change the world file permissions on Synology NAS systems. So we need to make sure that the default stays 0:0 and only if someone specifies PUID/PGID env vars we change to those. Speaking of which, let's name the env vars PUID/PGID as that's sort of the standard with most images.

Also let's do such a substantial PR against the `dev` branch instead of `main`. So we can incubate and test it with more people over a couple of weeks.

bootstrap
```diff
@@ -16,6 +17,29 @@ main() {
 }
 
 
+# Apply user id and group id
+apply_permissions() {
+    if [[ -n "$USER_ID" && -n "$GROUP_ID" ]]; then
```
lloesche, line 22: Instead of checking if these are defined let's put defaults into the `defaults` file.
Something like
```
PUID=${PUID:-0}
PGID=${PGID:-0}
```
and use the same defaults in the `Dockerfile`.

martinfelis: Sounds good to me. Please have a look at the updated branch.

lloesche:
👍 🚢

Issue #349, opened 2021-05-20
Add possibility to store non-zip Backups


TheHades:
I'm using [borgbackup](https://borgbackup.readthedocs.io/en/stable/) for creating backup, because it supports deduplication and compression.

How about adding an boolean option called "BACKUPS_ZIP" or "BACKUPS_COMPRESS" ?
If yes, then zip backups.
If false, then store "raw" backups of the worlds.

lloesche:
Any solution will have to be backwards compatible. Esp. take `PRE_BACKUP_HOOK` and `POST_BACKUP_HOOK` into consideration which replace the `@BACKUP_FILE@` string in the executed hook command.

What should a "raw" backup of the worlds look like? Just a copy of the worlds directory with a timestamp in the name?

TheHades:
> What should a "raw" backup of the worlds look like? Just a copy of the worlds directory with a timestamp in the name?

Yes, that should be sufficient.

I did a quick test with borgbackup (https://borgbackup.readthedocs.io/en/stable/index.html).

/tmp_valheim/backups = Contains unzipped Backups 
/tmp_borg = Borg Repository

Size of all unzipped Backups (25 backups * 60MB of Data * 2 (.db and .db.old file) = ~3GB)
```
root@server1 ~ # du -sh /tmp_valheim/
3.0G	/tmp_valheim/
```

```
# ls -al /tmp_valheim/backups/worlds-20210606-081010/worlds/
[...]
-rw-r--r-- 1 root root 61928780 Jun  6 10:01 'myworld.db'
-rw-r--r-- 1 root root 61928780 Jun  6 09:41 'myworld.db.old'
-rw-r--r-- 1 root root       49 Jun  6 10:01 'myworld.fwl'
-rw-r--r-- 1 root root       49 Jun  6 09:41 'myworld.fwl.old'
```

Now if we run a borgbackup on our "raw" backups, borg will compress those files AND it will detect duplicate blocks as well.

```
# borg info /tmp_borg/
Repository ID: f701065f547041e9ddfce587680f5ca86e1a592da93f7fb17ec0ccca1d0139cb
Location: /tmp_borg
Encrypted: No
Cache: /root/.cache/borg/f701065f547041e9ddfce587680f5ca86e1a592da93f7fb17ec0ccca1d0139cb
Security dir: /root/.config/borg/security/f701065f547041e9ddfce587680f5ca86e1a592da93f7fb17ec0ccca1d0139cb
[...]

PR #372, opened 2021-06-14
Add BACKUPS_ZIP environment variable and backup logic

TheHades:
See https://github.com/lloesche/valheim-server-docker/issues/349


valheim-backup
```diff
@@ -50,14 +50,27 @@ backup() {
...
+    # Wenn BACKUPS_ZIP true, dann zip
+    if [ ${BACKUPS_ZIP} = true ]; then
+        backup_file="$BACKUPS_DIRECTORY/worlds-$(date +%Y%m%d-%H%M%S).zip"
+        pre_backup_hook "$backup_file"
+        info "Backing up Valheim server worlds to $backup_file"
+        zip -r "$backup_file" "worlds/"
+        chmod "$BACKUPS_FILE_PERMISSIONS" "$backup_file"
+        post_backup_hook "$backup_file"
+    else
+        find worlds/ -name "${WORLD_NAME}.db" -type d | fatal "Could not find world file for world ${WORLD_NAME}"
```
lloesche, line 66: This should likely be an OR `||` not a pipe and also not `fatal`. Fatal will end the backup script. Meaning on a fresh install with no existing worlds the backup script would just error out instead of trying again later.

valheim-backup
```diff
@@ -50,14 +50,27 @@ backup() {
...
+        backup_file="$BACKUPS_DIRECTORY/${WORLD_NAME}-$(date +%Y%m%d-%H%M%S).db"
+        pre_backup_hook "$backup_file"
+        info "Backing up Valheim server worlds to $backup_file"
+        cp -v worlds/${WORLD_NAME}.db ${backup_file}
```
lloesche, line 70: It would make sense to also copy the backup db files as those are in a consistent state. This db file might be in the process of being written to when you copy it hence resulting in a broken backup file.

lloesche:
Are you still working on this?

valheim-backup
TheHades: Should we then just copy the .db.old file instead of the .db file, or both?

lloesche: I mean both likely makes sense. The .db file is newer so that's likely the one you'd want to restore from. However there's a slight chance it's corrupted if the backup runs mid-world-save, whereas the .old should always be in a consistent state.

lloesche (changes requested):
Can you fix the failing tests? Other than that it looks good! Once the CI reports all green we can merge this.

TheHades:
> Can you fix the failing tests? Other than that it looks good! Once the CI reports all green we can merge this.

Done.

lloesche (approved):

ibex-are-goats:
I'm testing this now and I'm still getting zips made. Anything I'm missing?

Latest version:
> Using default tag: latest
> latest: Pulling from lloesche/valheim-server
> Digest: sha256:e70a9f8968cc0b217dadb1110691ac601259ffda45b22db271ba4971ee9c188e
> Status: Image is up to date for lloesche/valheim-server:latest
> docker.io/lloesche/valheim-server:latest

In environment:
>      - BACKUPS_ZIP=false

Getting:
> worlds-20210828-044005.zip

Not sure how to double-check that the actual running container is the latest version, but I think I did recreate it right.

TheHades:
> I'm testing this now and I'm still getting zips made. Anything I'm missing?

Yes. As far as i can tell, the images on docker.io aren't updated anymore(?)
CC @lloesche 

You have to pull the image from https://github.com/lloesche/valheim-server-docker/pkgs/container/valheim-server.

PR #377, opened 2021-06-23
Ensure binary is executable before launch.

FennyFatal:
Running on Synology, I found that the exec flag was not being set by the steam utilities. Perhaps it should be elsewhere, but just before running seems safe enough to me.


lloesche:
That seems odd. We have a ton of Synology users. I feel like somebody would have stumbled across this.
I'm not opposed to adding this patch, it's low risk enough, I just wonder if the issue is somewhere else.

PR #384, opened 2021-08-09
Add an additional workaround for QNAP + ZFS

yostinso:
Per my own investigation and workaround discovery, this appears to be a common bug for both Steam & games that try to query disk space and get back too-large numbers for free space.


lloesche:
@yostinso thank you for the PR! LGTM 🚢

Issue #401, opened 2021-09-18
Synology - Doesn't work after new update


jaytark:
Hi,

Since the new updated container was released, I can no longer run the container.  In fact, it sometimes hoses my server and I have to restart the server.  I had to disable the ability to auto restart due to this.  It was working beautifully before the recent container update, even on Hearth & Home client.  I backed up the db and world files to another drive, and just deleted everything related to the docker image and reinstalled - same results.  

I've even uninstalled Docker, deleted all associated files and directories, reinstalled docker, and reinstalled/configured the latest Valheim container... still no success and same error.  Again, everything worked before the new container.

More Info:  On Synology, I'm pulling from Docker Hub (as I've always have) and pulling the 'latest'.  I'm not sure of a way to pull the old image back so we can continue to play as before.  It doesn't give an option to pull a previous version.

****Update:  Thumbing through real-time logs on Synology Docker, I came across this line:  "**_valheim-updater ERROR - Failed to download Valheim server from Steam - retrying later - check your networking and volume access permissions_**"

I've added firewall rule to Synology for ports 2456, 2457, although I've never needed to do that before... Problem still persists.  Any help or advise would be greatly appreciated.  Thanks in advance.

lloesche:
What update specifically are you referring to? The base functionality of the container hasn't changed since Jun 10th.
Dockerhub had stopped automatic builds for non-paying customers a while back. That should be fixed with yesterdays commit 2d2a9ad8e516a8451b5febdee62e9ab1986ea770 which builds on Github and then pushes to Dockerhub. But maybe there's a difference between the images Github and Dockerhub builds? I don't have a Synology so can't test it myself.

jaytark:
Whichever is the latest image on Dockerhub.  I pull "lloesche/valheim-server:latest".  I had pulled it late yesterday evening after everyone got off the server from playing to update it.  I only pulled it because your page said  "updated xx hours ago", so I assumed you updated something to the image based on the new Valheim update.  I turned off the active container, pulled lloesche/valheim-server:latest, reset the container, then started it back up.  Then the error is as posted above.

kraaijmakers:
I got the same problem, it sometimes freezes the whole Synology NAS.

slindebe:
The latest Valheim patch today changed something. IMHO it has nothing to do with the image.

I had HTOP running in a SSH window when my Synology became unresponsive, and I could see that a steamcmd process was taking all available CPU resources.

[...]

PR #407, opened 2021-09-20
Downgrade Debian 11 (stable) -> 10 Buster to debug #401

lloesche:


schildbach:
Just out of curiosity, can you say what's the reason for the downgrade?

PR #412, opened 2021-09-26
Build images once and push to both GHCR & Docker Hub

TheCatLady:
No need to have two workflows to publish images to both GHCR & Docker Hub. 😉


lloesche (approved):
I had no idea, this is awesome! 🚢

lloesche:
@TheCatLady btw. I saw that `docker/build-push-action@v2` supports Github Cache API directly using
```
                    cache-from: type=gha
                    cache-to: type=gha,mode=max
```

Would it make sense to migrate to that?

PR #417, opened 2021-10-05
Switch from tcpdump to ss for connection checks

### LINKED ISSUE #413: Idle check fails (says always idle) when SERVER_PUBLIC = 1 and PUID != 0
Some time ago, my server stopped creating backups except when restarting, saying the server is always idle. Now I've finally figured out why.

I have SERVER_PUBLIC=false and PUID=something-non-root, so the idle check uses `tcpdump`. The problem is that only root can do `tcpdump` and when it fails, the script assumes that the server is idle when it's actually not:

```
tcpdump: eth0: You don't have permission to capture on that device
(socket: Operation not permitted)
```

Log gives the following, even though I am on the server at that moment and have also been on the 29th.

```
Sep 30 23:00:09 supervisord: valheim-backup DEBUG - [71] - Received signal to backup world
Sep 30 23:00:09 supervisord: valheim-backup DEBUG - [71] - Not running backup as there has been no server activity since Tue 28 Sep 2021 10:32:04 PM UTC
```

We should at least make it so that when `tcpdump` fails for other reasons than "no packets", it is assumed that the server is NOT idle. I'll try to look into that, but I wanted to post this issue first.

solacelost:
This changes from tcpdumping with a 3 second timeout while looking for packets to checking active UDP sessions being tracked. This works completely unprivileged and I can confirm it working on rootless podman with SELinux enforcing on a Fedora system.

fixes #413


solacelost:
Also, a side note, it's possible that implementations differ based on kernel or something other than what I can think of, but I see UDP connections falling off of the `ss` list about 30 seconds after a client disconnects.

lloesche (dismissed):
lgtm, thank you! 🚢

lloesche:
The CI check failed with the following errors:

```
#37 3.203 In /usr/local/etc/valheim/common line 143:
#37 3.203         if [ $(ss -u | grep -F "$SERVER_PORT" | wc -l) -gt 0 ]; then
#37 3.203              ^-- SC2046: Quote this to prevent word splitting.
#37 3.203                        ^-- SC2126: Consider using grep -c instead of grep|wc -l.
```

Essentially it would like to see
```
if [ "$(ss -u | grep -cF "$SERVER_PORT")" -gt 0 ]; then
```

solacelost:
Fixed. I should have run shellcheck. `| wc -l` is just second nature to me :)

solacelost:
I'm unsure how it behaves in Docker with the default bridge networking there. As I mentioned, I'm running with rootless podman which means using slirp4netns networking. This is what I see:

```
[valheim@games ~]$ podman exec -it valheim2 bash
root@78e4cf38f876:/# ss -u # nobody is connected right now
Recv-Q                    Send-Q                                        Local Address:Port                                         Peer Address:Port
root@78e4cf38f876:/# ss -u # this is with one user logged on
Recv-Q                    Send-Q                                       Local Address:Port                                          Peer Address:Port
0                         0                                               10.0.2.100:40022                                           10.0.2.100:2456
```

solacelost:
UDP is stateless, but the implementation of tracking that I'm seeing in SS is unrelated to the state of the UDP connection in the same way that the TCP session would be tracked. I noted that the connections dropped off of my `ss` output after a pretty-strict 30 seconds. It's possible that this is related to the nftables implementation I have on my host system, outside of the container. If you're using an `iptables` based interface, it should be hitting similar netfilter interfaces in the kernel to affect the UDP tracking.

I just double-checked and see that when I run `sudo sysctl net.netfilter.nf_conntrack_udp_timeout` I get the output `net.netfilter.nf_conntrack_udp_timeout = 30`.

Could you check the value on your host system, @Tremolo4 ?

Tremolo4:
That seems to work for me too, thanks a lot!

I've confirmed that it does not pick up any other UDP traffic outside of the valheim container, not even from another valheim container.

One problem is that this method can't distinguish port numbers, so we can't check for game traffic on port 2456 only. I've been running the server in public mode to make the idle check work, so I very regularly get queries on port 2457 from people refreshing the server list. They get picked up by `tcpdump -n udp` and also by `nstat`. I've turned public mode off just now and there are also still some queries to port 2457 (without response this time), but a lot fewer. I guess these are from people whose cached server browser list still contains my server or other server monitoring tools.

The other non-player UDP packets I see in tcpdump are DNS queries from `steamcmd` whenever the update check runs.

Since even a single player on the server causes a lot more packets per second than this, I suggest we use a threshold to determine whether it's player traffic or not. Something like 30 packets per 3 seconds should be good. With just me on the server I'm getting around 100 - 400 packets per 3 seconds, depending on whether I move around or not etc. (using `docker exec -it <container_name> timeout 3 tcpdump -n udp and port 2456`).

Unless of course you know a way to check only a specific UDP port :D

solacelost:
Just made one other tweak because my `.vimrc` automatically prunes lots of whitespace when I save and I didn't catch that it slipped into the documentation commit, so I split the commit up into two chunks. Up to you if you'd like to have the idle detection documented, we can revert or I can drop it from the tree.

lloesche:
@solacelost looks good. But please revert the removal of trailing double spaces in the README. They are Markdown Syntax to enforce linefeeds.

solacelost:
Any update? I'm already running this version here successfully. Just hope the PRs can get cleaned up here so I can shift back to your image :)

PR #438, opened 2022-01-20
Update valheim-backup - add .fwl and .fwl.old

TheHades:
I think this fix should be merged ASAP.
The .fwl file contains the metadata for the Valheim world, so it is important to backup those file as well.


lloesche (approved):
thank you! lgtm 🚢

PR #475, opened 2022-05-05
can specify which Valheim plus release

Gyran:
if latest tag points to the wrong version, we can use the VALHEIM_PLUS_RELEASE env variable to specify which version to download instead of hardcoding it on the server. Or if for some reason you don't want the latest but pin a specific version.


InB4DevOps (approved):
lgtm

lloesche:
@Gyran thank you! 🚢

PR #485, opened 2022-06-17
Add clarification to *LIST_IDS environment variables

damianb:
Minor nit but hoping it saves someone else time down the road - the IDs need to be SteamID64 format which Valve documents here: https://developer.valvesoftware.com/wiki/SteamID

(Side note: today I learned there's a *third* SteamID format if steamid.io is correct. Thanks valve.)


mdPlusPlus:
So is it the SteamID64 hex format, e.g. `0x0110000100000000`?

damianb:
No, it's the 64-bit int.

PR #490, opened 2022-06-21
Fix for update 0.209.9

super-cooper:
With version 0.209.8, Valheim now saves worlds in `/config/worlds_local` rather than `/config/worlds`. I have attempted to patch the scripts to use this new path. I tested lightly, but there are likely to be some bugs.

### DIFF
--- a/common
+++ b/common
@@ -111,15 +111,16 @@ printline() {
 
 
 get_worlds_dir() {
-    WORLDS_DIR="/config/worlds_local"
+    worlds_dir="/config/worlds_local"
     if [ ! -d "${WORLDS_DIR}" ]; then
         if [ ! -d "/config/worlds" ]; then
             debug "No Valheim worlds to backup"
             return
         else
-            WORLDS_DIR="/config/worlds"
+            worlds_dir="/config/worlds"
         fi
     fi
+    return "$worlds_dir"
 }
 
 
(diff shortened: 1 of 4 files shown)


common
```diff
@@ -110,6 +110,20 @@ printline() {
 }
 
 
+get_worlds_dir() {
+    worlds_dir="/config/worlds_local"
+    if [ ! -d "${WORLDS_DIR}" ]; then
```
lloesche, line 115: Where is (capital) `$WORLDS_DIR` coming from at this point? From a global context or should it be the lower case one defined in the line above?

lloesche:
@super-cooper thank you for the PR, will review right away.

super-cooper:
I fixed the capitalization issue in `common`, but it is still capitalized in the backup script. I was unsure about the pattern with regards to which variables are capitalized. I can fix that one as well if desired.

DannyEss:
Since the latest update, the Server isn't accessable when using Valheim Plus as Env Variable, but it starts again using BepInEx as Env Variable. I already tried the updater script for both mods (valheim plus and bepinex), redeployed it, restarted it, but I always get the message "cant connect" when I try to play.
Valheim Plus Developer says it is BepInEx's fault, but I can access the Server again, when I turn off Valheim Plus but enable BepInEx
grafik

AdamLMark:
The fix for me was to copy my world(s) from /config/worlds to /config/worlds_local (had to overwrite files)

lloesche (approved):

PR #508, opened 2022-10-02
Add dependency for playfab plugin used by new -crossplay option

mmis1000:
Changes:

1. Add dependency required by the playfab plugin used by new -crossplay option. 
2. Bump debian version because the playfab plugin requires c and c++ lib version that isn't available in debain buster. 
3. Change lib32gcc1 to lib32gcc-s1  because it is renamed in the debian bullseye repository

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:buster-slim as build-env
+FROM debian:bullseye-slim as build-env
 ENV DEBIAN_FRONTEND=noninteractive
 ARG TESTS
 ARG SOURCE_COMMIT
@@ -78,7 +78,7 @@ RUN mkdir -p /usr/local/etc/supervisor/conf.d/ \
 RUN echo "${SOURCE_COMMIT:-unknown}" > /usr/local/etc/git-commit.HEAD
 
 
-FROM debian:buster-slim
+FROM debian:bullseye-slim
 ENV DEBIAN_FRONTEND=noninteractive
 COPY --from=build-env /usr/local/ /usr/local/
 COPY fake-supervisord /usr/bin/supervisord
@@ -92,7 +92,7 @@ RUN groupadd -g "${PGID:-0}" -o valheim \
     && apt-get -y --no-install-recommends install \
         libc6-dev \
         lib32stdc++6 \
-        lib32gcc1 \
+        lib32gcc-s1 \
         libsdl2-2.0-0 \
         libsdl2-2.0-0:i386 \
         cron \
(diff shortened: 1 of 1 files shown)


mmis1000:
It seems that. Besides this, there are more places broken. For example, cross play server don't use `$SERVER_PORT` at all. It onlu use  `$SERVER_PORT + 1`. And steam query returns 0 because it isn't on steam.

fr3sh68:
I built a docker image out of this and I can confirm the -crossplay option works and allows Xbox Game Pass PC version to connect, which it didn't previously.

I did get an error when building the docker image with line 165 in the Dockerfile with the range of ports: EXPOSE 2456-2457, I had to remove the range and just use EXPOSE 2456.  It might be because I built the image with podman, not docker.

mmis1000:
Not sure about that, I only have Ubuntu/Debian Machine. And they both have a quite recent kernel. So, no weird error about steamcmd so far. Probably need some synology user to find it out.

lloesche:
Are the newly added dependencies available on Debian 10? Could we add them without upgrading the entire OS?

fr3sh68:
I thought I read somewhere (Valheim Discord) that Debian 11 (Bullseye) was required.  I could be wrong though.  FYI, there's a new docker folder included in the download for Valheim Server 211.8 and it uses ubuntu:22.04

mmis1000:
> Are the newly added dependencies available on Debian 10? Could we add them without upgrading the entire OS?

No, c++ /c lib version on debian 10 is smaller than absolute minimum requirement.

lloesche (approved):
Ok well let's upgrade and see if Debian 11 works out well for our NAS users. It's been a while since #401 
If not we'll revert and create a separate image tag that has the crossplay support and one for older Kernels.

rkaput:
> Ok well let's upgrade and see if Debian 11 works out well for our NAS users. It's been a while since #401 If not we'll revert and create a separate image tag that has the crossplay support and one for older Kernels.

Synology user here. Server was working fine yesterday, but definitely does not work today after the upgrade. Happy to share any helpful info that I can if needed.

lloesche:
There is now a new Docker image `lloesche/valheim-server:debian10` which uses the older Debian 10 base image, but therefor doesn't support crossplay. This should be compatible with any NAS using old Kernels.

opello:
I wonder if it would be worth having a Synology user test Debian 12 (bookworm, glibc 2.36) with the deadlock problem?  It was easy enough to move the two `FROM debian:bullseye-slim` to `bookworm-slim` and it ran happily enough for me.  Maybe the newer glibc would work more nicely with Linux 3.10?

I came across a couple ([19329](https://sourceware.org/bugzilla/show_bug.cgi?id=19329), [28357](https://sourceware.org/bugzilla/show_bug.cgi?id=28357)) issues that are fixed after 2.29 (in bullseye) and related to thread start up issues.

Issue #515, opened 2022-11-05
Log spam "Waiting for server to listen on UDP port 2456" when using crossplay


Elk8324:
When crossplay is enabled (via adding "-crossplay" to SERVER_ARGS), the server process will not listen on the SERVER_PORT and will instead accept connections via PlayFab.

However the server listening check at https://github.com/lloesche/valheim-server-docker/blob/c94f0a50ea3b31a954a06642eca9db383c637723/valheim-server#L119 runs regardless, writing the above log message every 5 seconds.

Any suggestions for a clean fix?

SammyIsra:
I am also seeing this. Is this just a logging issue or a symptom of a larger game connection issue?

hancbrya:
<img width="1176" alt="Screenshot 2023-03-14 at 4 55 15 PM" src="https://user-images.githubusercontent.com/41393144/225159988-d9715ea7-16d1-4b9c-a1d8-66dbc56b747f.png">
I have the same issue, removing the `-e SERVER_ARGS="-crossplay"` arg resolves the issue. But unable to open port with cross play enabled.

PR #523, opened 2022-11-20
Fix server idle & server listening checks on crossplay-enabled servers

### LINKED ISSUE #521: valheim-status reports 0 players online with -crossplay, resulting in lack of backups
It seems like the player count isn't reported correctly (it's always 0) when crossplay is enabled, resulting in backups not being created due to "the server being idle".

https://github.com/lloesche/valheim-server-docker/blob/f47432bfcc9de4308cfe78346bbfa579dbe4317b/common#L155-L169

A potential temporary fix to get the backups working again is to set `SERVER_PUBLIC=false` which forces UDP packet counting logic.

It might be worth falling back to UDP packet counting when `-crossplay` flag is present unless there's a better solution.

recursiveGecko:
Resolves  #521, #515


foobarmeow (dismissed):
I'm no maintainer but LGTM!

serpi90 (approved):

lloesche:
Hey, thanks for the PR. Question, does this still work with servers that are not using crossplay and are not public? I remember vaguely that non-public servers only ever listen on 2456 and never bind the query ports.

wcdanh2:
+1, This looks like a good solution.

lloesche (approved):
Perfect, thank you @recursiveGecko

Issue #403, opened 2021-09-19
Waiting for server to listen on UDP port 2456


Aizuu:
Hello, running docker container on GCP. Seeing some issue with the server failing to start.
Seems to be some permissions issue, but the container seems to have access to the directories and has created some structure there:

```
root@e2665804626a:/opt/valheim/server# ls
 LinuxPlayer_s.debug   UnityPlayer_s.debug                    linux64           start_server_xterm.sh   steamclient.so          valheim_server_Data
 UnityPlayer.so       'Valheim Dedicated Server Manual.pdf'   start_server.sh   steam_appid.txt         valheim_server.x86_64
```

Container run:
```	docker run -d \
    --name valheim-server \
    --cap-add=sys_nice \
    --stop-timeout 120 \
    -p 2456-2457:2456-2457/udp \
    -v $HOME/valheim-server/config:/config:Z \
    -v $HOME/valheim-server/data:/opt/valheim:Z \
    -e SERVER_NAME="SANITIZED" \
    -e WORLD_NAME="valheim01" \
    -e SERVER_PASS="SANITIZED!" \
    lloesche/valheim-server
```

Server output:
```root@e2665804626a:~# valheim-server  
INFO - Running Valheim Server
DEBUG - [601] - Server config is name: valheim, port: 2456/udp, world: valheim01, public: 1, mod: none
DEBUG - [601] - Valheim server with PID 605 stopped
INFO - Releasing PID file /var/run/valheim/valheim-server.pid
INFO - Shutdown complete
root@e2665804626a:~# I0918 23:51:03.244550     609 main.go:76] Configuring Valheim server log filter
DEBUG - [601] - Waiting for server to listen on UDP port 2456
I0918 23:51:03.244981     609 main.go:96] Removing log lines starting with '(Filename:'
I0918 23:51:03.245212     609 main.go:89] Removing log lines matching ' '
I0918 23:51:03.245432     609 main.go:125] Removing invalid UTF-8 chars: true
I0918 23:51:03.245666     609 main.go:122] Removing empty log lines: true
I0918 23:51:03.245904     609 main.go:96] Removing log lines starting with 'src/steamnetworkingsockets/clientlib/steamnetworkingsockets_lowlevel.cpp'
/usr/local/bin/valheim-server: line 111: /opt/valheim/server/valheim_server.x86_64: Permission denied
DEBUG - [601] - Waiting for server to listen on UDP port 2456
[...]

PR #544, opened 2022-12-17
Use Debian 10 for the 32-bit environment

opello:
Some Synology users face a problem with glibc >= 2.31 due to collisions in the syscall number assignments.  Synology uses some that have since been used for other things in mainline Linux.  The one that exposed the issue here is clock_gettime64 being #403 in mainline Linux while in the Synology kernel exhibiting the issue it is SYNOArchiveBit.  Suffice it to say these functions are not interchangeable.  It seems that glibc can not tell the difference and working around the issue is required.

This seemed as clean as providing a wrapper library using LD_PRELOAD that ensured none of the 32-bit programs this image cared about would step on the wrong syscalls, because it is difficult to control what programs may call what other programs.  This is admittedly a little dirty, since the x86_64 execution will be using glibc 2.31 and the i386 environment will be using 2.28.

In practice I did not see any adverse effects, but it is not heavily exercised.  Just in launching valheim-server with -crossplay and ensuring libpary.so loaded successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,28 +78,37 @@ RUN mkdir -p /usr/local/etc/supervisor/conf.d/ \
 RUN echo "${SOURCE_COMMIT:-unknown}" > /usr/local/etc/git-commit.HEAD
 
 
+FROM --platform=linux/386 debian:buster-slim as i386-libs
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update \
+    && apt-get -y --no-install-recommends install \
+        libc6-dev \
+        libstdc++6 \
+        libsdl2-2.0-0 \
+        libcurl4 \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+
 FROM debian:bullseye-slim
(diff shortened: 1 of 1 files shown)


opello:
It might be most appropriate to just push this as a special synology tagged build with some explanation about it being for their special kernel if you have the "steamcmd takes 99% CPU" issue?

lloesche:
Hey @opello, thank you for the PR!

This is pretty funky indeed, but if it means that we can ditch the debian10 branch and provide crossplay for all I'm all for it.

> it is difficult to control what programs may call what other programs

I'm not really concerned with any programs outside of valheim-server and the misc. helpers this container uses. So as long as it's compatible with those I think this is good to merge to main.

lloesche (approved):

opello:
You're quite welcome, thanks for the container in the first place!

All of the other pieces that I saw were 64-bit binaries anyway, so this shouldn't even affect them.  The big 32-bit binary that was a problem was steamcmd and this does avoid the large issue there on Synology hardware and doesn't introduce any issues that I saw on vanilla kernels.

PR #556, opened 2023-01-05
Fix rebooting when players are online.

wcdanh2:
When the cron job for the daily reboot launches valheim_is_idle which subsequently launches valheim-status, valheim-status does not have the environment variables it needs and falls back to defaults. On servers that are not using the default ports this causes the status to timeout which is treated as 0 players online resulting in a reboot even when players are online.

This patch modifies the crontab file so that the cron environment will have SERVER_PORT and STATUS_HTTP_HTDOCS which are the 2 parameters that valheim-status tries to get from the environment.

### DIFF
--- a/valheim-bootstrap
+++ b/valheim-bootstrap
@@ -80,6 +80,8 @@ write_permittedlist
 
 # Create crontabs
 crontab=$(mktemp)
+echo "SERVER_PORT=$SERVER_PORT" >>"$crontab"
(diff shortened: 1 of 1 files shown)


wcdanh2:
I haven't gotten the pr modified yet but, I found that "SERVER_PUBLIC" also needs to be added.

I'll make a new pr if this is already merged (hard to see on my phone)

Issue #551, opened 2023-01-01
Restart if idle only restarts when NOT idle


ajgoda90:
If RESTART_IF_IDLE is set, it checks for idle, but then restarts when there are players. valheim-is-idle returns 0 if idle, but the restart is triggered if it returns 1. I noticed this because I forgot to set a timezone so it was happening at 11PM instead of 5AM while we were playing.

`# valheim-is-idle returns 0 if the server is idle or 1 if players are connected`
[https://github.com/lloesche/valheim-server-docker/blob/249cf9572f1c58fb13b87b81806cf760a5dd4094/valheim-is-idle#L2](valheim-is-idle)

`echo "$RESTART_CRON $cmd_valheim_is_idle && $cmd_supervisorctl restart valheim-server" >> "$crontab"`
[https://github.com/lloesche/valheim-server-docker/blob/d1e7298f8021dec926dc106d8d26adffc3ba6303/valheim-bootstrap#L96](valheim-bootstrap)

wcdanh2:
`
server_is_idle() {
    if [ "$SERVER_PUBLIC" = 1 ]; then
        "$cmd_valheim_status" > /dev/null 2>&1
    else
        # Throw away datagram statistics since last run
        _udp_datagram_count &>/dev/null
        # Wait to track datagrams over window
        sleep "$IDLE_DATAGRAM_WINDOW"
        if [ "$(_udp_datagram_count)" -gt "$IDLE_DATAGRAM_MAX_COUNT" ]; then
            return 1
        else
            return 0
        fi
    fi
`

I may be reading this wrong since bash isn't my strongest language, but it seems that server_is_idle() does not have a return value if the server is public.

wcdanh2:
I'm also having the issue of my servers restarting at midnight EST (5am UTC) but after some testing I'm thinking it may not be the cron job.

To test it I opened a shell in my docker image like so: `docker exec -ti valheim-server /bin/bash` then in the shell ran the following: `/usr/local/bin/valheim-is-idle && echo "restart"`

When there is a player on the server the `echo "restart"` does not get evaluated, when there is no one on the server it prints "restart"

wcdanh2:
Narrowing this down a bit, It appears that valheim-status does not execute properly in the cron environment and is getting a timeout error and returning 0.

CrypticSafe5:
[...]

PR #560, opened 2023-01-10
Add SERVER_PUBLIC to cron environment

wcdanh2:
When SERVER_PUBLIC is set to false the server does not respond on the query port.  When server_is_idle() is run under cron environment SERVER_PUBLIC is unset and defaults to true causing the function to attempt to use the query port.

This patch adds SERVER_PUBLIC to the cron environment so that server_is_idle() will use the correct path for determining if players are online.

See #551 Restart if idle only restarts when NOT idle.

### DIFF
--- a/valheim-bootstrap
+++ b/valheim-bootstrap
@@ -80,8 +80,11 @@ write_permittedlist
 
 # Create crontabs
 crontab=$(mktemp)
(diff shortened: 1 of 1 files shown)


SolidBrave:
So is this resolved because for me the server still just restarts even if players are connected?

Issue #568, opened 2023-01-18
Valheim Plus without passwort - environment variable SERVER_PASS can't be leaved blank


JackTheFragger:
To run a local dedicated server on my synology nas I wanted to disable the password for valheim plus.

I've edited the valheim_plus.cfg as you described. The variable VALHEIM_PLUS is set to "true" and SERVER_PUBLIC to "false". But the environment variable SERVER_PASS can't be set to empty ("") because you have to set a value. By deleting the value SERVER_PASS is your default value "secret".

That's why it isn't possible to run a lan dedicated server without password.

Possible solutions could be to not have a default password or to use an environment variable like "SERVER_DISABLE_PASS" to "true" with default value "false" to set an empty password.

caseodilla:
Are you deleting the entire variable `SERVER_PASS`, or are you just setting `SERVER_PASS=`, i.e. setting it to blank?

JackTheFragger:
I've tried both. If I delete the entire variable, the password will be 'secret' because it's the default value. Setting 'SERVER_PASS=' isn't possible because the GUI of docker ap at my synology nas doesn't allow blank values.

PR #572, opened 2023-01-26
Setting SERVER_PASS to empty value when VPLUS PW is disabled

JackTheFragger:
Fix for Issue #568

Setting SERVER_PASS to empty value if variable 'VPCFG_Server_disableServerPassword' is true


lloesche (changes requested):
shellcheck is complaining about accessing `VPCFG_Server_disableServerPassword` without defining it.

You could reduce the code to something like:

```
SERVER_PASS=${SERVER_PASS-secret}
if [ "${VPCFG_Server_disableServerPassword:-}" = true ]; then
	SERVER_PASS=""
fi
```

lloesche:
@JackTheFragger thank you for the PR! I added a comment to the code above.

JackTheFragger:
> shellcheck is complaining about accessing `VPCFG_Server_disableServerPassword` without defining it.
> 
> You could reduce the code to something like:
> 
> ```
> SERVER_PASS=${SERVER_PASS-secret}
> if [ "${VPCFG_Server_disableServerPassword:-}" = true ]; then
> 	SERVER_PASS=""
> fi
> ```

Where do I define it?

lloesche:
In `defaults`. At the same point where you made the current changes.

PR #578, opened 2023-01-31
Added a section in the variables about crossplay

falkoro:


falkoro:
need to change it to the correct args.

PR #601, opened 2023-03-24
Edited infos for ports and LAN-Play

JackTheFragger:
- added remark for port forwarding with crossplay enabled
- added remark for LAN-Play only possible with crossplay disabled
- added remark for port configuration for mods


lloesche (approved):
@JackTheFragger thank you, this is great! 🚢 👍

PR #654, opened 2024-01-22
Add param to rsync for mods that will remove unused mods

SavedByGrace251:
This will ensure a strict sync between the config plugins folder and the server's plugins folder.  That way if a mod is deleted in the config folder, it will also be removed from the server.


madmarxofficial:
Great, I just have the same idea.
Would be great to remove the zombies config files too but we would need a way to know that the deleted mod created it in the first place.

lloesche (approved):
@SavedByGrace251 thank you for your PR! lgtm 🚢

Issue #759, opened 2025-10-02
PERMISSIONS_UMASK fails when set to 027


Xeboc:
When the umask is set to 027, the math for the `DEFAULT_FILE_PERMISSIONS` fails:

```bash
PERMISSIONS_UMASK=027

echo $(printf "%#o" $(( 8#0666-8#$PERMISSIONS_UMASK )))
0637
```

Instead of subtracting, this would work using a bitwise AND with inverted umask:

```bash
PERMISSIONS_UMASK=027

echo $(printf "%#o" $((0666 & ~8#$PERMISSIONS_UMASK)))
0640

echo $(printf "%#o" $((0777 & ~8#$PERMISSIONS_UMASK)))
0750
```

```bash
PERMISSIONS_UMASK=022

echo $(printf "%#o" $((0666 & ~8#$PERMISSIONS_UMASK)))
0644

echo $(printf "%#o" $((0777 & ~8#$PERMISSIONS_UMASK)))
0755
```

PR #760, opened 2025-10-02
Update UMASK calculation

Xeboc:
- Fixes the math for the UMASK calculations.
- Drops the 8# in front of 0666 and 0777 because bash already treats a number starting with 0 as an octal.

Closes: https://github.com/lloesche/valheim-server-docker/issues/759


lloesche:
Hey @Xeboc thank you for this PR! LGTM 🚢

PR #777, opened 2026-03-30
Add Thunderstore mod management via vmm

endoze:
Manual mod installation is error-prone and requires server downtime to update. By integrating the Valheim Mod Manager (vmm) directly into the container, mods from Thunderstore can be defined in a simple list and kept in sync automatically on the same schedule as server updates.

Users can now specify a mod list via MOD_MANAGER_MODS or a config file, and the container handles downloading, installing, updating, and pruning stale mods without manual intervention. Patcher-type mods (e.g. HookGenPatcher) are handled transparently by copying patcher assemblies into BepInEx's patchers directory. The server is restarted automatically whenever mod state changes.


MrSmiley801:
This would be a welcomed change! Manual mod management via SFTP or whatever other means is tedious and no fun.

PR #789, opened 2026-09-04
Modernize the valheim image

stevbev:
The commits are split by the concept to make the image more maintainable. While I have an opinion about all the changes, the desired commits can be selectively accepted for enhancement instead.

- Optimize image layers by in-lining DEBIAN_FRONTEND variable.
- Use native Debian busybox, golang, supervisor, tox packages
- Modernize i386 package installation
- Use steamcmd:debian-trixie as base image to get steam


Shawak:
This drops version pinning for a bunch of stuff, not a fan.

stevbev:
> This drops version pinning for a bunch of stuff, not a fan.

I am unsure what this comment is supposed to represent as it is very terse. I will do my best to engage on the topic with the statement as given.

This PR actually does not "drop version pinning for a bunch of stuff". What is does is shift where the pinning occurs to eliminate a few anti-pattern concepts introduced a long time ago. I made this PR to leverage engineering best practices to improve reliability and security in a way that decreases the maintainers future maintenance efforts. 
- The base image is using Debian Trixie, which is well known to pin package versions to specific upstream versions and backport security fixes. Shifting to use the Debian Trixie package that is actively maintained by a security team provides stability, predictability, and reliability. This means that generally, the packages that are relied upon only need extensive testing when changing the base image's major release and eliminates the need to get the legacy software running on a modern distribution.
- The static defined busybox 1.36.1 has 4 medium severity vulnerabilities from 2023 that would not be an issue today if the base image's busybox package was used. A few of the optional busybox commands were replaced with the native distribution commands to avoid having to unnecessarily manage a custom busybox build and installation. There are no components in the Valheim image that require pinning to version 1.36.1, so shifting to the native distribution package provides a modern and security covered dependency.
- supervisor 4.2.5 is now using the native 4.3.0 package. There are no components in the Valheim image that require pinning to version 4.2.5, so shifting to the native distribution package provides a modern and security covered dependency.
- golang 1.24.1 has 1 critical, 2 high, 1 medium severity vulnerabilities that would not be an issue today if the base image's golang package was used. The go program does not have any version requirement that requires remaining on the unmaintained 1.24 branch, so shifting to the native distribution package provides a modern and security covered dependency.
- Replacing the Debian Buster i386 libs with the Trixie i386 libs resolves a (potential) compatibility problem that was introduced during the migration to Trixie as the new base image.

I am happy to discuss further why you aren't a fan if you can articulate the opinion. Cheers!

Shawak:
You are now building from steamcmd/steamcmd:debian-trixie which is a moving target and has no pinned BusyBox, Go or Python version. These version may change since steamcmd/steamcmd:debian-trixie points to latest.  I am not saying the current versions are good, its just harder to maintain if its unknown what version someone is using when submitting bug tickets.

JustinTArthur:
Nothing intrinsically wrong about any of these and modernization is generally a good thing. I'm going to decline it for a few reasons:
* There's a planned overhaul to the busybox side, leaning into it for more like crond, upgrading (like this PR), but also only compiling the minimum set of required busybox tools.
* Supervisor upgrade is good, it'll be nice to retire setuptools, but this will continue to be frozen to a PyPI version for consistency across even distro upgrades.
* The i386 modernization has been done a couple times but usually gets reverted. There's nothing wrong with the better way, it's just that freezing to that anceint glibc allows SteamCMD to continue to work on a few ancient NAS/appliance kernels. If SteamCMD doesn't get a 64-bit release soon, we may do something like this anyway and offer an alternative Dockerfile for NASes.

Supervisord, busybox, and maybe cron updates could land within a week or so. I'm keeping image basics stable this week so that for our Valheim 1.0 infrastructure support, the only thing we need to worry about is the 1.0 changes (such as for zone-chunked world DB).

Thanks for putting in effort to help make the image better, sorry for not taking it in this time.

stevbev:
> You are now building from steamcmd/steamcmd:debian-trixie which is a moving target and has no pinned BusyBox, Go or Python version. These version may change since steamcmd/steamcmd:debian-trixie points to latest. I am not saying the current versions are good, its just harder to maintain if its unknown what version someone is using when submitting bug tickets.

This comment leads me to believe you are not very familiar with Debian package policy or the policies any of the other non-rolling Linux distribution. The steamcmd/steamcmd:debian-trixie image is using Debian Trixie as a base image. Debian Trixie is not, as you say, a moving target. Debian Trixie has pinned busybox 1.37.0,  there is no sliding release involved. Same thing with most of the other packages installed on Debian releases. You can verify this yourself by looking at Debian's releases over the last 20 years to get a sense of the stability of the Debian package pinning policy. 

Not knowing which version of busybox is a red herring; all users are generally running the legacy version on the older image, or the new version the team selects to use in the future. If the exact version of busybox causes such consternation, investing the effort to move to busybox alternatives would be a superior solution than staying on an ancient version of a package with known security vulnerabilities that cannot be changed ever because of reasons.

There also seems to be a gross misunderstanding of what steamcmd is and does. Steam releases steamcmd/steamcmd images based on many different Linux distributions as the base image. Steam installs and configures the Steam client in a manner that makes it easy for end users to obtain support from the distribution community (if distribution problem), or from Steam (if Steam problem). There is a "moving target" here in that the image is updated with the latest Steam client, and the security updates installed when the base distribution updates their image. Good clean and modern CI/CD pipeline designed to make maintainers lives easier. Steam is not installing new versions of your favorite tools on the distribution image, they are the same packages everyone gets by using the base distribution image.


Anyway, bringing facts to an opinion argument isn't particularly effective, so I will stop here. I appreciate the past efforts the maintainers have put into designing, publishing, and supporting this image. My intent was to give back to the community that made it easy for me to stand up a server to play with friends. I did my part by spending a lot of time improving the architecture and submitting the change upstream, and now I can move on.

PR #794, opened 2026-09-08
Fix DEBUG_START_FRESH cleanup

Shawak:
Fixes `DEBUG_START_FRESH` not properly clearing `/opt/valheim`.

The glob was quoted, preventing it from being expanded and leaving the existing server data untouched.

### DIFF
--- a/valheim-bootstrap
+++ b/valheim-bootstrap
@@ -21,7 +21,7 @@ fi
 # check if we are supposed to wipe our data directories
 if [ "$DEBUG_START_FRESH" = true ]; then
     warn "Wiping all data directories (DEBUG_START_FRESH: $DEBUG_START_FRESH)"
-    rm -rf "/opt/valheim/*"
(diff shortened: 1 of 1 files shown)


valheim-bootstrap
```diff
@@ -21,7 +21,7 @@ fi
 # check if we are supposed to wipe our data directories
 if [ "$DEBUG_START_FRESH" = true ]; then
     warn "Wiping all data directories (DEBUG_START_FRESH: $DEBUG_START_FRESH)"
-    rm -rf "/opt/valheim/*"
+    rm -rf /opt/valheim/*
```
JustinTArthur, line 24: Seems fine. Maybe add a `mkdir -p "$STATUS_HTTP_HTDOCS"` afterwards in case they have custom htdocs? I'm guessing that got missed when this was first written (alongside the bad glob-quoting)

Issue #773, opened 2026-03-03
BACKUPS_IF_IDLE=false prevents automated backups: player check fails completely (both with and without crossplay)


TorstenAden:
Bug Description:
When setting BACKUPS_IF_IDLE=false, the automated backups never trigger. The backup script incorrectly assumes the server is empty (0 players) and skips the backup, even when players are actively playing. This happens regardless of whether crossplay is enabled or disabled.

Environment:
- Image: lloesche/valheim-server
- Platform: Ugreen NAS (Docker)
- Network Mode: Host 
- Game clients: PC

The Root Cause:
The backup script relies on checking network sockets (UDP 2456) to determine if players are online. This method no longer works reliably for modern Valheim connections in this environment. The socket check returns 0 active connections, which causes the cron job to abort.

The server log clearly registers the connections correctly:
"Player joined server "ServerName" that has join code XXXXXX, now 1 player(s)"

Expected Behavior:
The backup script should accurately detect connected players. Since checking network ports is no longer reliable for Valheim, a robust fix would be to have the script parse the valheim-server log for the current player count, rather than relying on socket checks.

TorstenAden:
Update: The player detection issue goes both ways, making the idle-check completely unusable.

Overnight, while the server was completely empty (0 players logged in), the automated system continuously logged:
"Players connected to Valheim server - skipping update check"

Because the script falsely detected players, the BACKUPS_IF_IDLE=false condition never triggered. This resulted in backups running every 30 minutes all night long on an empty server.

It appears the current network socket check (UDP 2456/2457) cannot distinguish between actual player connections and standard background traffic (such as continuous A2S_INFO server browser pings from Steam master servers). 

Relying on network socket monitoring for player detection is broken in both directions (false negatives during active Crossplay/PlayFab sessions, and false positives from server pings when empty). Parsing the actual server log for player count is strongly recommended to resolve this completely.

PR #808, opened 2026-09-12
fix: detect CROSSPLAY=true in idle checks

### LINKED ISSUE #807: Broken player detection under Crossplay (PlayFab) breaks BACKUPS_IF_IDLE and causes constant SteamCMD lag spikes via UPDATE_CRON
Bug Description:
When CROSSPLAY=true is enabled, the container's internal player detection mechanism completely fails to detect active players. The container continuously evaluates the server as having 0 connected players.

This causes two severe cascading issues:
1. BACKUPS_IF_IDLE=false prevents all scheduled world backups. Once BACKUPS_IDLE_GRACE_PERIOD expires after boot, the backup daemon aborts every run with "Not running backup as there has been no server activity since [boot timestamp]", even while multiple players are actively building and saving chunks.
2. UPDATE_IF_IDLE=true fails to protect active sessions. Because the updater always evaluates "No players connected to Valheim server", UPDATE_CRON (default: */15 * * * *) triggers steamcmd verification every 15 minutes while players are in-game, causing recurring I/O and CPU lag spikes.

Environment:
- Image: lloesche/valheim-server:latest (ghcr.io/community-valheim-tools/valheim-server:latest)
- Game Version: Valheim 1.0 (network version 40)
- Network Mode: Host
- Configuration: CROSSPLAY=true, BACKUPS_IF_IDLE=false, UPDATE_IF_IDLE=true

Log Evidence:
Server log confirms active players joined via PlayFab:
"Player joined server "Dagcenter_Ugglan" that has join code XXXXXX, now 1 player(s)"
"World save (...) done."

Yet valheim-updater continuously reports:
"DEBUG - [35] - No players connected to Valheim server"
"Downloading/updating/validating Valheim server from Steam"

And valheim-backup reports:
"DEBUG - [75] - Not running backup as there has been no server activity since Sat Sep 12 03:00:45 AM CEST 2026"

Expected Behavior:
The idle/player detection script (valheim-is-idle) should monitor the server's stdout/log for PlayFab player connect/disconnect events or check chunk modification timestamps under worlds_local/<WorldName>/ rather than relying solely on UDP socket checks that do not capture PlayFab traffic.

llit47:
## Problem

When `CROSSPLAY=true`, the launcher adds `-crossplay` to the Valheim process, but `server_is_crossplay()` inspected only `SERVER_ARGS`. A normally configured crossplay server with no `-crossplay` in `SERVER_ARGS` was therefore classified as non-crossplay by `server_is_idle()`.

In a real runtime reproduction with one connected player, `server_is_crossplay()` returned false. `valheim-status` took the A2S path, timed out, and exited 0; `server_is_idle()` consequently returned true while the player was online. This PR does not change that A2S timeout/exit-code behavior.

## Root Cause

Both `CROSSPLAY=true` and manually passing `-crossplay` through `SERVER_ARGS` enable crossplay, but detection recognized only the latter.

## Fix

`server_is_crossplay()` now returns true when `CROSSPLAY=true` or `SERVER_ARGS` contains `-crossplay`. The existing case-insensitive `SERVER_ARGS` behavior is preserved.

## Testing

- Focused regression tests pass; against base `1c9a6ea`, they fail on `CROSSPLAY=true`. Manual `SERVER_ARGS=-crossplay` and uppercase `-CROSSPLAY` compatibility cases pass.
- `bash -n` and `git diff --check` pass. Docker image build and the Dockerfile ShellCheck stage passed.
- Real runtime acceptance test on Valheim 1.0.12 with crossplay/PlayFab, `CROSSPLAY=true`, and no `-crossplay` in `SERVER_ARGS` (patched image commit `3acd1646c7bc967c170b3c05a543e03634ddeff4`):
  - 0 players: `server_is_crossplay=true`; UDP sample `0`; `server_is_idle=true`.
  - 1 connected AFK player: ten 3-second UDP samples `204, 201, 208, 205, 206, 196, 206, 203, 204, 211`; all BUSY; `server_is_idle=false`.
  - 1 active player: ten 3-second UDP samples `208, 262, 245, 243, 240, 247, 232, 246, 252, 255`; all BUSY; `server_is_idle=false`.
  - After disconnect: ten 3-second UDP samples `0, 1, 0, 0, 2, 2, 0, 0, 1, 1`; all IDLE; `server_is_idle=true`.

This runtime test validates the existing `nstat` crossplay path in the tested environment; it does not establish that the UDP heuristic is universally reliable.

## Scope

No changes to `nstat` behavior, `valheim-status`, PlayFab implementation, updater/restart behavior, or backups. Related context: #773; this PR does not fully resolve it.

### DIFF
--- a/valheim-tests
+++ b/valheim-tests
@@ -67,6 +67,4 @@ main() {
 }
 
 test_server_is_crossplay || exit 1
-if [ "${1:-}" != "--crossplay-only" ]; then
-    main
-fi
+main
--- a/valheim-tests
+++ b/valheim-tests
@@ -2,6 +2,26 @@
 . /usr/local/etc/valheim/defaults
 . /usr/local/etc/valheim/common
 
+test_server_is_crossplay() {
+    if ! (CROSSPLAY=true; SERVER_ARGS="-saveinterval 1800"; server_is_crossplay); then
+        echo "CROSSPLAY=true should enable crossplay detection"
+        return 1
+    fi
+    if ! (CROSSPLAY=false; SERVER_ARGS="-saveinterval 1800 -crossplay"; server_is_crossplay); then
+        echo "SERVER_ARGS=-crossplay should enable crossplay detection"
+        return 1
+    fi
+    if ! (CROSSPLAY=false; SERVER_ARGS="-saveinterval 1800 -CROSSPLAY"; server_is_crossplay); then
+        echo "SERVER_ARGS=-CROSSPLAY should enable crossplay detection"
+        return 1
+    fi
+    if (CROSSPLAY=false; SERVER_ARGS="-saveinterval 1800"; server_is_crossplay); then
+        echo "Crossplay detection should be disabled without CROSSPLAY=true or -crossplay"
+        return 1
+    fi
+    echo "Crossplay detection tests passed"
+}
+
 main() {
     timeout=900
     echo "Running tests..."
@@ -46,4 +66,7 @@ main() {
     exit 0
 }
 
-main
+test_server_is_crossplay || exit 1
+if [ "${1:-}" != "--crossplay-only" ]; then
+    main
+fi
--- a/common
+++ b/common
@@ -161,7 +161,7 @@ _udp_datagram_count() {
 
 
 server_is_crossplay() {
-    [[ "${SERVER_ARGS,,}" =~ "-crossplay" ]];
+    [[ "$CROSSPLAY" = true || "${SERVER_ARGS,,}" =~ "-crossplay" ]]
 }
 
 


valheim-tests
```diff
@@ -46,4 +66,7 @@ main() {
     exit 0
 }
 
-main
+test_server_is_crossplay || exit 1
+if [ "${1:-}" != "--crossplay-only" ]; then
+    main
+fi
```
JustinTArthur: Was this intended to be left in? If so, note we're already using `$1` for the name via matrix.

If you meant to remove, go ahead and remove it and I'll merge.

JustinTArthur (changes requested):

JustinTArthur (approved):

valheim-tests
llit47: Yep, that was only intended for the focused test path. I've removed it in `6293e3f` so `$1` remains available for the existing matrix name. The crossplay regression tests still pass. Thanks!

TorstenAden:
Regarding the UDP packet heuristic via nstat:

In environments where the container runs with `network_mode: "host"` (which is standard practice on NAS setups like UGOS, Synology, or Unraid to handle port mapping cleanly), the `nstat` check monitors the entire host network namespace rather than an isolated container socket.

On an active NAS with 0 players connected and the Valheim server completely idle, routine background UDP traffic (mDNS, UPnP, NTP, local discovery) consistently trips the counter. A 3-second sample on an empty server logged 64 incoming UDP datagrams (`UdpInDatagrams`), causing `valheim-is-idle` to continuously return `1` (busy / not idle).

As a result, `server_is_idle` never transitions to `true`, keeping `BACKUPS_IF_IDLE=false` permanently locked in an active state.

Relying on raw UDP datagram counters is fundamentally unreliable when `network_mode: "host"` is used. A persistent fix should check internal game events instead, such as:
1. Parsing server stdout/log output for player join/disconnect events (PlayFab or Steam).
2. Monitoring file modification timestamps on active chunk/save files in `worlds_local/<WorldName>/` (for Valheim 1.0+).

Issue #529, opened 2022-12-05
crontab entries broken when uid and gid are defined in valheim.env


pixel8ed:
The bootstrap script fails to create crontab entries due to missing permissions if you set the UID and GID for the valheim user.

`valheim.env` parameters defined:
```
PUID=1000
GUID=1000
```

Error example:
```
valheim@a950e2c141e2:~$ crontab -l
crontabs/valheim/: fopen: Permission denied
```

Temporary fix:
in `valheim.env`:
```
PRE_SUPERVISOR_HOOK=/usr/sbin/usermod -a -G crontab valheim
```

Fix example:
```
valheim@5e7699d3b6cb:~$ crontab -l
0 * * * * [ -f "/var/run/valheim/valheim-backup.pid" ] && kill -HUP $(cat /var/run/valheim/valheim-backup.pid)
*/15 * * * * [ -f "/var/run/valheim/valheim-updater.pid" ] && kill -HUP $(cat /var/run/valheim/valheim-updater.pid)
0 5 * * * /usr/local/bin/valheim-is-idle && /usr/local/bin/supervisorctl restart valheim-server
```

lloesche:
I'm trying to reproduce the issue but so far can't. Maybe you can see where I'm going wrong:

I run a test server using
```
[lukas@blackhole ~]$ docker run --rm -it -e PUID=1000 -e PGID=1000 lloesche/valheim-server
INFO - Setting uid:gid of valheim to 1000:1000
INFO - Setting timezone Etc/UTC
INFO - Setting up syslogd - logging to stdout
2022-12-22 02:27:22,510 INFO Included extra file "/usr/local/etc/supervisor/conf.d/syslog.conf" during parsing
2022-12-22 02:27:22,510 INFO Set uid to user 0 succeeded
2022-12-22 02:27:22,512 INFO RPC interface 'supervisor' initialized
2022-12-22 02:27:22,512 INFO supervisord started with pid 1
2022-12-22 02:27:23,513 INFO spawned: 'syslogd' with pid 24
2022-12-22 02:27:23,514 INFO spawned: 'crond' with pid 25
2022-12-22 02:27:23,515 INFO spawned: 'valheim-bootstrap' with pid 26
Dec 22 02:27:23 syslogd started: BusyBox v1.34.1
2022-12-22 02:27:23,515 INFO success: valheim-bootstrap entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
[...]

PR #809, opened 2026-09-12
Install valheim's crontab from bootstrap so cron works under no-new-privileges

### LINKED ISSUE #805: Cron doesn't seem to work in 1.2.0
It seems since https://github.com/community-valheim-tools/valheim-server-docker/commit/c900868d6bc3f80470f98a4ce6747988fe94d565 that cron no longer works, at least with PUID/PGID set. I noticed this as the valheim-updater didn't work, but also backup isn't working. Relevant log lines:
```
2026-09-11 21:45:19.639 info 2026-09-11 21:45:19,639 INFO success: crond entered RUNNING state, process has stayed up for > than 1 seconds (startsecs) 
2026-09-11 21:45:18.367 unk Sep 11 21:45:18 supervisord: valheim-bootstrap crontab: must be suid to work properly 
2026-09-11 21:45:18.366 debug Sep 11 21:45:18 supervisord: valheim-bootstrap DEBUG - [25] - Creating cron to restart valheim-server using schedule 10 5 * * * 
2026-09-11 21:45:18.366 debug Sep 11 21:45:18 supervisord: valheim-bootstrap DEBUG - [25] - Creating cron to check for updates using schedule */15 * * * * 
2026-09-11 21:45:18.366 debug Sep 11 21:45:18 supervisord: valheim-bootstrap DEBUG - [25] - Creating cron to do world backups using schedule 5 * * * * 
2026-09-11 21:45:18.354 unk Sep 11 21:45:18 crond[24]: crond (busybox 1.36.1) started, log level 8 
2026-09-11 21:45:18.349 info 2026-09-11 21:45:18,349 INFO spawned: 'crond' with pid 24 
```

Perhaps it's the removal of `usermod -a -G crontab valheim`? I can look into this later if you don't have time, just let me know so we don't both work on it.

For reference I also set
```yaml
    cap_add:
      - sys_nice
    security_opt:
      - no-new-privileges:true
```

JustinTArthur:
Since v1.2.0, scheduled backups, update checks and restarts silently stop for anyone running with a non-zero `PUID` under a runtime that ignores setuid bits. The only symptom is one line at startup:

`valheim-bootstrap crontab: must be suid to work properly`

`crond` runs normally with no crontab, so nothing else looks wrong. Users are likely to discover it when they reach for a backup that was never taken.

Fixes #805.

## Root cause

c900868 replaced Debian's cron with busybox's. Because busybox's `crontab` applet is `BB_SUID_REQUIRE`, that commit added `chmod a+s /usr/local/bin/crontab` and dropped `usermod -a -G crontab valheim`.

`valheim-bootstrap` runs as `valheim` (`supervisord.conf`), so installing a crontab now depended on that setuid bit. `no-new-privileges`, Kubernetes' `allowPrivilegeEscalation: false` (required by the restricted Pod Security Standard, and set by OpenShift's `restricted-v2` SCC) and `nosuid` mounts all make the kernel ignore it, leaving euid non-zero, and busybox refuses. `PUID` defaults to 0, where the check is skipped — which is why this wasn't caught.

Pre-1.2.0 was not relying on privilege escalation at all: Debian's `crontab` is *setgid*, which `no-new-privileges` ignores equally, and it worked only because `valheim` belonged to the `crontab` group (`3c75737`, added for #529 — the 2022 report of this same failure) and the spool is group-writable.

Two side effects of that `chmod a+s` are also fixed here. The path is a symlink, so the bit landed on the shared busybox binary itself; and the line inserted itself ahead of three continuation lines, so `steamcmd`, `steamerrorreporter` and `/usr/bin/supervisord` were silently made setuid/setgid instead of receiving the intended `755`.

## Change

`bootstrap` already runs as root before supervisord, so it now builds and installs the crontab there, and `valheim-bootstrap` no longer does. It's fed via stdin because busybox opens a file argument as the target user, who can't read root's temp file. A failed install is now logged as an error rather than passing unnoticed. `chmod a+s` is gone, so busybox is no longer setuid.

`supervisord.conf` is untouched — `crond` keeps its existing slot, command and `autostart`.

**Behaviour note:** `PRE_BOOTSTRAP_HOOK` can no longer influence `UPDATE_CRON`/`BACKUPS_CRON`/`RESTART_CRON`, since generation now happens earlier. That was never documented — it worked only because the hook is `eval`'d in the same shell. `PRE_SUPERVISOR_HOOK` can still do it.

## Testing

Verified under rootless Podman on x86_64 by layering the branch's files onto the published `1.2.0` image (v1.2.0 is this branch's base), confirming in each case that `crond` actually executed a job as `valheim`, not merely that a crontab file appeared:

| Scenario | Result |
|---|---|
| `PUID=1000` + `no-new-privileges` | crontab installed root-owned, job ran as uid 1000 |
| `PUID=0` + `no-new-privileges` | installed, job ran as uid 0 |
| `PUID=1000`, no flag | installed, job ran |
| Schedule set only via `PRE_BOOTSTRAP_HOOK` | not applied — expected, see note above |
| Install failure (spool directory removed) | `ERROR - Failed to install crontab …`; updater and backup keep running |
| Reproduction on unmodified `1.2.0` | `must be suid to work properly`, empty spool |
| Baseline `1.1.0` | works, via `crontab` group membership |
| `1.1.0` with `valheim` removed from that group | fails under the flag, confirming the setgid path is ignored |

`shellcheck` passes with the project's flags.

## Related

Builds on #785, which creates `/var/run/valheim` at runtime.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -185,7 +185,6 @@ RUN groupadd -g "${PGID:-0}" -o valheim \
     && chown -R valheim:valheim /var/run/valheim \
     && chown -R root:root /opt/steamcmd \
     && chmod u=rwx,go=rx /opt/steamcmd/steamcmd.sh \
-    && chmod a+s /usr/local/bin/crontab \
     /opt/steamcmd/linux32/steamcmd \
     /opt/steamcmd/linux32/steamerrorreporter \
     /usr/bin/supervisord \
--- a/bootstrap
+++ b/bootstrap
@@ -13,6 +13,7 @@ main() {
     setup_supervisor_http_server
     setup_status_http_server
     pre_supervisor_hook
+    setup_crontab
     exec /usr/bin/tini -- /usr/local/bin/supervisord -c /usr/local/etc/supervisord.conf
 }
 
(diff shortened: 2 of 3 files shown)


cosandr (approved):
Seems to work, crontab is being loaded now.

PR #799, opened 2026-09-09
Recover from inaccessible installed Steam depot manifests

OwenBinchy:
SteamCMD can get stuck requesting an inaccessible installed depot manifest, for example after leaving `public-test`.

This adds one retry with `validate` when a fresh content-log entry reports that specific denial for an installed manifest and Supervisor reports the server as `STOPPED`. Configured SteamCMD arguments are preserved. The old app manifest is backed up; the backup directory prevents further metadata resets until manually cleared, even after successful recovery.

Failed downloads no longer overwrite installed files or merge mods from a partial cache. Fallback startup checks the selected vanilla/modded installation.

This is a targeted workaround, not a general fix for SteamCMD update failures. Detection depends on SteamCMD's log wording: if that changes, recovery will not trigger, but normal update attempts continue.

Related: #795, #758.

Issue #795, opened 2026-09-09
Server not updating to 1.0


Andargor:
My client upgraded fine through steam. I shut down and removed the docker instance, erased all worlds and started up a new instance. It still reports 0.221.12 after update.

I am seeing the following messages in docker logs:

Sep  9 13:45:10 supervisord: valheim-updater
Sep  9 13:45:10 supervisord: valheim-updater Connecting anonymously to Steam Public...^[[0m
Sep  9 13:45:11 supervisord: valheim-updater OK
Sep  9 13:45:11 supervisord: valheim-updater ^[[0mWaiting for client config...^[[0m
Sep  9 13:45:11 supervisord: valheim-updater OK
Sep  9 13:45:11 supervisord: valheim-updater ^[[0m
Sep  9 13:45:11 supervisord: valheim-updater Waiting for user info...^[[0mOK
Sep  9 13:45:11 supervisord: valheim-updater ^[[0m
Sep  9 13:45:13 supervisord: valheim-updater  Update state (0x0) unknown, progress: 0.00 (0 / 0)
Sep  9 13:45:13 supervisord: valheim-updater ^[[0m
Sep  9 13:45:13 supervisord: valheim-updater Error! App '896660' state is 0x6 after update job.
Sep  9 13:45:13 supervisord: valheim-updater ^[[0m
Sep  9 13:45:13 supervisord: valheim-updater Unloading Steam API...^[[0m
Sep  9 13:45:14 supervisord: valheim-updater OK
Sep  9 13:45:14 supervisord: valheim-updater ^[[0m
Sep  9 13:45:14 supervisord: valheim-updater ERROR - Failed to update Valheim server from Steam - however an existing version was found locally - using it
Sep  9 13:45:14 supervisord: valheim-updater .d..t...... ./
Sep  9 13:45:14 supervisord: valheim-updater INFO - Valheim Server is already the latest version

Andargor:
Update: after completely deleting all files and making a new instance, the new version is installed.

However, I am concerned that future automated updates may be impacted.

sbalgas:
I ran into the same issue when switching from `public-test` back to `public`.

I already had:

```env
PUBLIC_TEST=false
STEAMCMD_ARGS=validate -beta public
```

but SteamCMD was still failing with:

```text
Error! App '896660' state is 0x6 after update job.
```

Checking:

```bash
tail -100 /home/valheim/Steam/logs/content_log.txt
```

[...]

Issue #713, opened 2024-11-28
Outdated version of server


wildBcat:
I freshly pulled the image for this server but when I try to load into it, it says that the server version is older than the game version. If the updater script running correctly on this?

geeklynad:
Having the same issue. Seems to be a problem with SteamCMD.

```
valheim-updater Error! App '896660' state is 0x6 after update job.
supervisord: valheim-updater ERROR - Failed to update Valheim server from Steam - however an existing version was found locally - using it
```

This is on an older container that hasn't been run in a while. I tried to bring it back up today, but it has been unable to update.
Tried going into the container and manually uninstalling/reinstalling...
`cd /opt/steamcmd`
`./steamcmd.sh +force_install_dir /opt/valheim +login anonymous +app_uninstall 896660 +quit`
`./steamcmd.sh +force_install_dir /opt/valheim +login anonymous +app_update 896660 validate +quit`
but no dice. The update appears to succeed, but the installation is still an older version. I might have just had the locations wrong or something.

What seems to have worked was to delete the data volume from outside container, while it was stopped and removed, and then reinitialize the container from scratch, but still using the config folder with world data and whatnot.

wildBcat:
Hmm, I have not been able to get that to work. I have also tried deleting all the folders related to it and doing a fresh install and I still seem to get an outdated version.

TH3-5L0TH:
> What seems to have worked was to delete the data volume from outside container, while it was stopped and removed, and then reinitialize the container from scratch, but still using the config folder with world data and whatnot.

This worked for me. Took a few minutes for the rebuilt container to download Valheim again and start accepting connections but updated clients can now join.

Docker compose down (Stops & removes the container essentially)
Delete contents of folder mapped to "/opt/valheim"
Docker compose up

geeklynad:
> Hmm, I have not been able to get that to work. I have also tried deleting all the folders related to it and doing a fresh install and I still seem to get an outdated version.

Have you tried creating an entirely new container from scratch? It sounds like what you did *should* have accomplished this, there's a possibility that there may have been some remnants that prevented reinitializing from scratch.

If an entirely new container works, just migrate your world data and plugins over to the new one.

wildBcat:
[...]

PR #812, opened 2026-09-13
Fix updating older installations on new start, survive Steam refusing prior manifest (state 0x6)

### LINKED ISSUE #758: No longer updating to the latest Valheim server version
Clients are getting "Incompatible version", and server displays:

```
valheim_1  | Sep 20 16:06:41 supervisord: valheim-server 09/20/2025 16:06:41: Network version check, their:35, mine:34 
valheim_1  | Sep 20 16:06:41 supervisord: valheim-server 09/20/2025 16:06:41: Peer 76561198034208212 has incompatible version, mine:0.220.5 (network version 34)   remote 0.221.4 (network version 35)
```

Already restarted the server, can't see anything obviously wrong in its console messages. 

TIA!

JustinTArthur:
Fixes #758 (and its duplicates #795, #713). Includes #799 by @OwenBinchy, cherry-picked with attribution.

## Cause

To update an existing installation SteamCMD loads the manifest of the installed build as well as the new one. It looks in `$HOME/Steam/depotcache` first and only asks Steam for a manifest request code on a cache miss. Anonymous logins are refused request codes for any manifest that is no longer the branch's current one, so the request fails with `Access Denied` and the job ends in state 0x6.

`$HOME` is not on a volume, so every container re-creation between two Valheim updates lost the cache while the stale appmanifest survived in `/opt/valheim/dl`. Process uid and file ownership make no difference; this was verified with root and `PUID=1000` on root-owned and chowned volumes, and confirmed by reading `steamclient.so`, which has no uid or ownership checks in the content path.

## Changes

- `valheim-bootstrap` symlinks `$HOME/Steam/depotcache` to `/opt/valheim/dl/depotcache` so the cache survives re-creation.
- `valheim-updater` (from #799) detects the denial for an installed manifest in the fresh part of the content log, moves `appmanifest_896660.acf` aside and retries with `validate`. SteamCMD then verifies the existing files and downloads only changed chunks. Failed downloads no longer rsync a partial cache over the installation.
- Follow-up on #799: recovery is no longer limited to a stopped server or to a single attempt; the old appmanifest is kept as `appmanifest_896660.acf.denied`.

## Testing

Built image, `PUID=1000`, root-owned `/opt/valheim` volume:

- Stale volume at build 25185644 without cache: 0x6, automatic recovery, 317 MB downloaded, server updated.
- Re-created container on the updated volume: already the latest version.
- Old build with the persisted cache: 38 MB delta update, no recovery needed.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>


valheim-bootstrap
```diff
@@ -49,6 +49,18 @@ fi
...
 
+# Keep SteamCMD's depot manifest cache on the /opt/valheim volume. SteamCMD
+# needs the installed build's manifest to update, and anonymous logins cannot
+# re-download a manifest once it is no longer the branch's current one, so
+# losing $HOME/Steam/depotcache on container re-creation breaks updates.
+mkdir -p "$steamcmd_depotcache_path" "$steamcmd_home"
+if [ -d "$steamcmd_home/depotcache" ] && [ ! -L "$steamcmd_home/depotcache" ]; then
+    cp -an "$steamcmd_home/depotcache/." "$steamcmd_depotcache_path/" 2> /dev/null || true
+    rm -rf "$steamcmd_home/depotcache"
+fi
```
OwenBinchy, line 72: Could we only delete the original cache if the copy succeeds? If the destination is full, this would silently lose the cached manifests.

OwenBinchy (approved):
Looks good - thanks for taking this further.

valheim-bootstrap
JustinTArthur: ef654cb433fa750581ec3947a5f313b9503862b4 ought to cover this now for most failures. Sorry about the verbose AI commit message.

